PR #173, opened 2017-05-05
node.js stores alias name, group.js reads alias if prop doesnt exist

Dsan10s:


LeaVerou:
Same here. You seem to have reverted some changes to Mavo due to using an older build.

LeaVerou:
@Dsan10s any updates?

Issue #42, opened 2016-03-11
Enter should insert item if in collection


LeaVerou:
Enter should insert item below (except if there's a default handling, like in textareas etc)

PR #175, opened 2017-05-13
Enter inserts item below current item and focuses on it

francesca-cicileo:
Issue: https://github.com/mavoweb/mavo/issues/42


LeaVerou:
Thanks, merged!

FWIW, `this.walkUp()` does return the right value, but I forgot to return that in `closestItem`. A `return` should've been sufficient, but this works too.

PR #200, opened 2017-05-22
Fix #180: Construct Date using local time if no timezone given

### LINKED ISSUE #180: time.datetime is shifted by one day
I'm using Firefox 53.0.2 under Linux, and the demo http://mavo.io/demos/foodie/ is buggy:
the displayed date is one day sooner than the stored date.

E.g. the default value is "2016-03-15", but "14 Mar 2016" is displayed.

ashsearle:
Fixes issue #180

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -608,19 +608,26 @@ function toDate(date) {
 		// Remove all whitespace
 		date = date.replace(/\s+/g, "");
 
-		// If no timezone, insert local
 		var timezone = Mavo.match(date, /[+-]\d{2}:?\d{2}|Z$/);
-
-		if (!timezone) {
-			var local = _.localTimezone;
-			var minutes = Math.abs(local % 60);
-			var hours = (Math.abs(local) - minutes) / 60;
-			var sign = local < 0? "-" : "+";
-			date += sign + twodigits(hours) + ":" + twodigits(minutes);
+		if (timezone) {
+			// parse as ISO format
+			date = new Date(date);
+		}
+		else {
+			// construct date in local timezone
+			var fields = date.match(/\d+/g);
+			date = new Date(
+				// year, month, date,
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hey, thanks for contributing!
Nice solution, but it's not that simple :(
This works only when you have an actual date, that starts with a year.
You might only have a time (and in the future we want to support more of the formats described in [the HTML5 spec](https://html.spec.whatwg.org/multipage/semantics.html#the-time-element)).
Thanks for looking into it!

ashsearle:
Hi Lea,

A few lines up are two pre-existing code blocks.  The first prefixes the time today's date if the string doesn't start with a date.  The second appends a time, if the time component doesn't exist.

My change is simply to use the multi-field Date constructor (which operates in local time), rather than using today's local timezone to try and form a fully qualified date+time+zone ISO string.

Sidenote:
You cannot assume today's timezone is the same as that timezone in effect for the given date.  You also need to know the time, and even then you still don't know the timezone.

e.g. In the UK 2016-10-30T01:30 (without timezone information) could represent either of:
Sun Oct 30 2016 01:30:00 GMT+0000 (GMT)
Sun Oct 30 2016 01:30:00 GMT+0100 (BST)

LeaVerou:
Oh I see! Yeah, that makes sense. (in general, date handling in Mavo is a bit of a hack, by large because date handling in browsers is a bit of a hack 😛)

Have you tested that the date-related tests still pass?

http://test.mavo.io/functions.html#dates
http://test.mavo.io/property.html#datetime-formatting

You can find instructions on how to test this here: https://github.com/mavoweb/test/blob/master/README.md

ashsearle:
I hadn't (I'd been wondering where the tests were...)

I've run the now.  All are still passing. :-)

LeaVerou:
Awesome, thank you so much!!

PR #202, opened 2017-05-24
fix #191

### LINKED ISSUE #191: Property can't be set when using the minified version
When using `mavo.js`, everything works fine, but when using `mavo.min.js` and entering Edit mode, the content of a property is cleared upon hover and I can't type anything.

Reloading the page shows only the last character I've typed.

Same results with **Firefox 53** and **Chrome 58**.

# My code

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <link rel="stylesheet" href="https://get.mavo.io/mavo.css">
    <!-- Works fine with https://get.mavo.io/mavo.js -->
    <script src="https://get.mavo.io/mavo.min.js"></script>
    <title>Mavo test</title>
  </head>
  <body>
    <div mv-app="test" mv-storage="local" mv-autosave="0">
      <div>
        <article property="element" mv-multiple>
            <h1 property="title">Title</h1>
            <span property="content">Content</span>
        </article>
      </div>
    </div>
  </body>
</html>
```

abo3desa:
I have fixed #191. Please let me know if there is anything I can do to make this successfully merged into master.


LeaVerou:
Thank you so much!!
I’m not a huge fan of transpiling twice, but it'll do for now.
The main issue is Stretchy: It seems you have replaced it with the minified version, but it's still listed as `stretchy.js` in the dependencies, so the wrong file would get copied in the future.
I can merge and fix it locally (since this is very high priority) or you could fix it. I'll wait a bit :)

LeaVerou:
Merged, thanks!

PR #229, opened 2017-06-27
[POC] hijacking addEventListener VS handleEvent

WebReflection:
This PR is a **proof of concept** and it should not be merged.

Having a `handleEvent` method in the primitive as such:
```js
handleEvent(evt) {
  this.edit.handleEvent.call(this, evt);
}
```
could be the only indirection needed to [avoid hijacking addEventListener](https://github.com/LeaVerou/bliss/blob/52b8429b049bbd5872db979dddbf997611f44677/bliss._.js#L56) to keep track of all the runtime created arrow functions.

Classes can define a single `handleEvent` method and the instance itself can be the only reference needed, and also always known, to remove any sort of listener.

The result is a less greedily patched environment, less memory usage, and every browser from IE9 compatible with such pattern (plus eventually [even IE8](http://webreflection.github.io/ie8/) through a little CDN based polyfill that won't affect any other browser).

Please note the pattern suggested in here was, again, a quick proof of concept, and you should feel free to improve it as much as you want.

Please also note in those cases where you need a closure the bliss solution might still be needed, or you should store that reference to the current context instead and avoid arrow functions on the wild, passing always the element itself, or [it's single delegated](https://medium.com/@WebReflection/dom-handleevent-a-cross-platform-standard-since-year-2000-5bf17287fd38#381b), as listener.

Best Regards

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -349,11 +349,12 @@ _.register({
 			if (this.attribute === "mv-clicked") {
 				element.setAttribute("mv-clicked", "0");
 
-				element.addEventListener("click", evt => {
-					let clicked = +element.getAttribute("mv-clicked") || 0;
-					this.value = ++clicked;
-				});
+				element.addEventListener("click", this);
 			}
+		},
+		handleEvent: function (evt) {
+			let clicked = +evt.currentTarget.getAttribute("mv-clicked") || 0;
+			this.value = ++clicked;
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Ahem, you still have event classes in your code. Also, `elements.js` is basically plugins on the core functionality of `primitive.js`. Rewriting the plugins does little…

WebReflection:
This is the whole point Lea
screenshot from 2017-06-27 22-27-37

You invoke edit with instances you already hold as context. There is no need in those `add/removeEventListener` to hold anything else.

Again, it's a proof of concept, you can figure out if you can migrate to a similar approach or keep hijacking the platform, your call.

LeaVerou:
What I'm saying is `mousemove.mavo:edit` is nonsensical to a browser without hijacking `addEventListener`, so your code wouldn't work.
Also, `primitive.js` is adding its own `.mavo:edit` handlers.
It's very easy to rewrite a part of a large project with code that doesn't work and proclaim "There! It's simple! That's the whole point, don't you see it??". The devil is usually in the details :)

WebReflection:
> mousemove.mavo:edit is nonsensical to a browser without hijacking addEventListener

I'm not sure I udnerstand.

```js
document.body.addEventListener('mousemove.mavo:edit', console.log);

document.body.dispatchEvent(new Event('mousemove.mavo:edit'));
```

You can use `handleEvent` in there too.

If it's about bringing magic to those types then any indirection would be both boring but also solving the issue.

LeaVerou:
Let me try once more: These events are not synthetic. `mousemove.mavo:edit` is just `mousemove` and is fired like a normal mousemove, the rest is a class to unbind it later. Are you suggesting they should become synthetic? Why? Your whole point is that classes are not needed, why not remove them then?

WebReflection:
> the rest is a class to unbind it later.

I thought these were meant to be event name, didn't know about their purpose.

> Your whole point is that classes are not needed, why not remove them then?

I guess you could, but you would need an indirection if you'd like to group events by "_class name_".

Anyway, my whole point is that *maybe* a `handleEvent` based EventListener here could avoid hijacking and the usage of an expando property per node to be able to remove listeners later on.

However, since I guess Bliss brings in here much more than just that, it might be pointless to even think about migrating 'cause Bliss, as it is, worked just fine so far.

As summary: this was just a proof of concept, something to eventually discuss/consider or ignore, if unable to solve anything.

Do what's best for this project.

Cheers

PR #238, opened 2017-07-10
Allow URL hash with leading integer

### LINKED ISSUE #220: Url fragment/hash with leading integer
Create a page with Mavo, index.html, then try to load index.html#1. This gives an error. #a1 or #a works fine, while #1a fails as well, so it's the leading integer that's the problem.

```
Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#1a' is not a valid selector.
    at https://get.mavo.io/mavo.js:1:987
    at HTMLHtmlElement.document.documentElement.addEventListener.evt (https://get.mavo.io/mavo.js:1365:25)
    at HTMLHeadingElement.fire (https://get.mavo.io/mavo.js:1:6185)
    at Object.function.n.type.r.element.n.Element.n.Element.(anonymous function) (https://get.mavo.io/mavo.js:1:8470)
    at Function.function.n.type.r.$.n.(anonymous function) [as fire] (https://get.mavo.io/mavo.js:1:8858)
    at e.dataChanged (https://get.mavo.io/mavo.js:2746:5)
    at e.dataChanged (https://get.mavo.io/mavo.js:3931:33)
    at sneak (https://get.mavo.io/mavo.js:3923:10)
    at e.sneak (https://get.mavo.io/mavo.js:1216:15)
    at Function.sneak (https://get.mavo.io/mavo.js:1233:31)
```

RubenVerborgh:
Fixes #220, in which using a hash URL starting with an integer would trigger an error.

The extra function might be overkill, but it seemed nicer than having `document.getElementById(location.hash.substr(1))` twice in a row.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -325,6 +325,14 @@ var _ = $.extend(Mavo, {
 	},
 
 	/**
+	 * Get the element identified by the URL hash
+	 */
+	getTarget: function() {
+		var id = location.hash.substr(1);
+		return document.getElementById(id);
+	},
+
+	/**
 	 * Object utilities
 	 */
 
@@ -620,12 +628,12 @@ document.addEventListener("blur", evt => {
 }, true);
 
 addEventListener("hashchange", evt => {
-	updateWithin("target", $(location.hash));
+	updateWithin("target", _.getTarget());
 });
 
 document.documentElement.addEventListener("mavo:datachange", evt => {
 	// TODO debounce
-	updateWithin("target", $(location.hash));
+	updateWithin("target", _.getTarget());
 });
 
 updateWithin("focus", document.activeElement !== document.body? document.activeElement : null);


LeaVerou:
Why not just 
```js
$(`[id="${location.hash.substr(1)}"]`)
```
?

RubenVerborgh:
Seems overkill to use Bliss for simple ID-based lookup, but I don't mind either way. Will update.

RubenVerborgh:
Ah no, here's a better reason: injection. Funny IDs would do interesting things. Given that we are already fixing an edge case here (numerical IDs), better to directly fix it for all IDs.

LeaVerou:
> Seems overkill to use Bliss for simple ID-based lookup

What’s overkill? Bliss just calls `querySelector`. It's not jQuery. These 3 lines are the whole implementation of `$()`: https://github.com/LeaVerou/bliss/blob/gh-pages/bliss.shy.js#L91-L95

> Ah no, here's a better reason: injection. Funny IDs would do interesting things.

Ok, that makes sense. Could you please add a `getTarget` function in util.js and use that? If we’re gonna have a helper function for this, it should be accessible to the rest of Mavo, it may be needed elsewhere at some point.

RubenVerborgh:
OK, added `getTarget` function.

PR #242, opened 2017-07-17
Add missing var declarations

valterkraemer:
Added missing variable declarations, execution crashed for me. (When I build Mavo with Webpack)


LeaVerou:
Ouch, thanks!

PR #243, opened 2017-07-17
Add "ignore newline character" to inline-svg:s

valterkraemer:
It produced an invalid CSS file for me when using Webpack. I tried to figure out why but couldn't (both were using the same version of node-sass).

However, to my understanding you are not allowed to use line-breaks in SCSS without escaping that character.


LeaVerou:
Thanks!

henrahmagix:
I also just noticed this! But I was installing Mavo via npm so I got v0.1.0, not v0.1.2 with the fix. @LeaVerou Please may we have the latest releases published to npm? I'm trying to use Mavo with webpack =)

I researched and logged an issue with sass-loader since node-sass didn't have a problem without this pull-request's changes: https://github.com/webpack-contrib/sass-loader/issues/490

PR #244, opened 2017-07-17
improved .eslintrc file with globals and indentation rules

valterkraemer:
These additions removed most of the invalid errors eslint gave me.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,6 +32,14 @@
 		"space-before-blocks": [1, "always"],
 		"arrow-spacing": 1,
 		"comma-spacing": 1,
-		"keyword-spacing": 1
+		"keyword-spacing": 1,
+		"indent": ["error", "tab", {
+			"outerIIFEBody": 0
+		}]
+	},
+	"globals": {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hi there,

Thanks for contributing. Where were you getting errors from eslint? I don't get any. Perhaps you were running it against the files in `dist/`?

valterkraemer:
I investigated a bit. The reason was that I have another .eslintrc file higher up in the folder hierarchy with different rules (in this case `"extends": "eslint:recommended"`) that were inherited to the Mavo project.

henrahmagix:
Adding `"root": true` will ensure eslint never combines with a developer's own config, say in their home directory (usually an ancestor directory to repos): https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy

henrahmagix:
Oh, just realised and read that configs in `~/.eslintrc.*` are ignored if project- or folder-level configs are found. That's a nice, somewhat expected feature. `"root": true` may still prove useful, but otherwise please ignore me =D

PR #247, opened 2017-07-18
Do not force user to log in on store

### LINKED ISSUE #245: Saving forces login even if not needed by backend
I'm working on a backend plugin. There are scenarios when you could, but don't need to log in to store data. Now Mavo forces me to always log in if it's possible to.

https://github.com/mavoweb/mavo/blob/master/src/mavo.js#L488

valterkraemer:
Fix #245


LeaVerou:
Thanks!

PR #249, opened 2017-07-20
Add missing stub for put on backend

RubenVerborgh:
`this.put` is called from `store`, but it didn't have a stub yet so fails now with `this.put is not a function`.

Alternatively, I can implement `put` with a rejected promise, or implement it as `HTTP PUT` (analogous to how `get` implements `GET`).

### DIFF
--- a/src/backend.js
+++ b/src/backend.js
@@ -49,6 +49,7 @@ var _ = Mavo.Backend = $.Class({
 	ready: Promise.resolve(),
 	login: () => Promise.resolve(),
 	logout: () => Promise.resolve(),
+	put: () => Promise.resolve(),
 
 	isAuthenticated: function() {
 		return !!this.accessToken;


LeaVerou:
Please check if this breaks the Remote backend.
I think right now there is no Save if `mv-storage` is set to a random URL because the Remote backend has no put. 

In any case, given that `put` is only called in one place, I'd be more inclined to add a check if it exists there, than to falsely report a resolved (or rejected) promise.
The other methods with stubs are called all over the place.

RubenVerborgh:
Sorry, which Remote backend? Is this a plugin?

> falsely report a resolved (or rejected) promise.

Well, if `put` is not implemented, the save failed, so reject seems fair.

LeaVerou:
> Sorry, which Remote backend? Is this a plugin?

That's the default backend, if the value of `mv-storage` is unrecognized.

RubenVerborgh:
OK. Remote does not seem to break, since it only sets `read` permissions, so `store` (and hence, `put`), will never be called.

LeaVerou:
Ok, but since it needs to be a rejected promise, I'm closing this.

PR #272, opened 2017-08-21
Update primitive value even if it's undefined

valterkraemer:
Fix for https://github.com/mavoweb/mavo/issues/273

Not sure if it causes some other problems


LeaVerou:
Hi there, thanks for contributing.

Not sure why you’re checking `this.modes`, especially twice?

valterkraemer:
Only checking it once. Not sure why needed to be checked at all. That was how it was done earlier.

PR #280, opened 2017-09-17
Adding locale.fr.js

Falafel65:
Yay, French !

I don't know if it's the right way to add translation, but anyway, here is the french one.


Falafel65:
Noooes I failed again at adding things to Mavo, this is obviously a plugin T_T
Closing here, opening at the plugins part.

LeaVerou:
Looks good apart from that!

LeaVerou:
Are you going to add it to plugins?

You can follow the Greek one as an example. Name it `mavo-locale-fr.js`.

Falafel65:
I intended to, then my Internet crashed 😞 

I made a repo (https://github.com/Falafel65/mavo-locale-fr), I was about to adding it via plugins.mavo.io, is that ok ?

LeaVerou:
Yes, that works great!

PR #308, opened 2017-11-20
allow line breaks in js expressions

### LINKED ISSUE #307: js expression with leading newline
leads to early return.
This works:
```
[some_property
/* js */
]
```
But this does not (at least in es5 bundle):
```
[
some_property
/* js */
]
```
because the compiled function 
```js
return
some_property
```
will have a `return` on a line by itself.
Here is a simple test:

no newline: https://codepen.io/anon/pen/XzVLzZ?editors=1000
newline: https://codepen.io/anon/pen/rYpEpJ?editors=1000

andreasplesch:
Allow line breaks in js mode expressions by enclosing in parens.
fixes #307

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -333,7 +333,7 @@ var _ = Mavo.Script = {
 
 		return new Function("data", `with(Mavo.Functions._Trap)
 				with (data || {}) {
-					return ${code};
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thank you!

PR #349, opened 2018-06-15
Add "contains" Mavo function

efeichen:


src/functions.js
```diff
@@ -324,6 +324,20 @@ var _ = Mavo.Functions = {
 
 	len: text => str(text).length,
 
+	/**
+	 * Return true if the group or collection has the needle (text) and false if not
+	 */
+	contains: (haystack, needle) => {
+		if ($.type(haystack) === "array" || $.type(haystack) === "object") {
```
LeaVerou: Not array! That's what we discussed yesterday, arrays will be handled separately.

efeichen: The discussion with David? Correct me if I'm wrong, so for arrays, you would return an array of booleans?

LeaVerou: Yup. You don't need to code that separately, you can use `Mavo.Script.binaryOperation()` (which is also used for operators, hence the name). It takes two params and if one of them is an array, it performs the operation you define via the third param (`o.scalar`) element-wise.

efeichen: Ok, probably gonna work on it on Monday though, plus a new permission denied error starts appearing now I think after I updated node again, so frustrating >:(

LeaVerou: No worries, have a good weekend, see you on Monday!

efeichen:
Had a problem with my local repository so I reforked on Github. A continuation of this PR is [here](https://github.com/mavoweb/mavo/pull/354)

src/functions.js
efeichen: I don't feel like I'm understanding the third param really well. Are you saying that if I pass in haystack and needle for scalar, if one of them is an array it will do the looping of its elements for me without me calling a for loop of some sort?

I tried to look through the function declaration in mavoscript.js but I'm not fully understanding what it's doing.

PR #354, opened 2018-06-18
contains(): Return an array of booleans when array passed

efeichen:
Add logic so an array of booleans would be returned if haystack of type array is passed to the function (w/o been called by another `contains()`).

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -361,7 +361,11 @@ var _ = Mavo.Functions = {
 	/**
 	 * Case insensitive search
 	 */
-	search: (haystack, needle) => haystack && needle? str(haystack).toLowerCase().indexOf((needle + "").toLowerCase()) : -1,
+	search: (haystack, needle) => {
+		haystack = str(haystack);
+		needle = str(needle);
+		return haystack && needle? haystack.toLowerCase().indexOf((needle).toLowerCase()) : -1;
+	},
 
 	starts: (haystack, needle) => _.search(str(haystack), str(needle)) === 0,
 	ends: function(haystack, needle) {
--- a/src/functions.js
+++ b/src/functions.js
@@ -332,8 +332,7 @@ var _ = Mavo.Functions = {
     contains: (haystack, needle) => {
 		var ret = Mavo.Script.binaryOperation(haystack, needle, {
 			scalar: (haystack, needle) => {
-				var type = $.type(haystack);
-				if (type === "object") {
+				if ($.type(haystack) === "object") {
 					for (var property in haystack) {
 						ret = _.contains(haystack[property], needle);
 						if (Array.isArray(ret)) {
@@ -351,10 +350,12 @@ var _ = Mavo.Functions = {
 			},
 			identity: null
 		});
+
 		// if result is an empty array, return false
 		if (ret.length === 0) {
 			return false;
 		}
+
 		return ret;
     },
 
@@ -364,7 +365,7 @@ var _ = Mavo.Functions = {
 	search: (haystack, needle) => {
 		haystack = str(haystack);
 		needle = str(needle);
-		return haystack && needle? haystack.toLowerCase().indexOf((needle).toLowerCase()) : -1;
+		return haystack && needle? haystack.toLowerCase().indexOf(needle.toLowerCase()) : -1;
 	},
 
 	starts: (haystack, needle) => _.search(str(haystack), str(needle)) === 0,
--- a/src/functions.js
+++ b/src/functions.js
@@ -330,9 +330,9 @@ var _ = Mavo.Functions = {
 	 * @returns Array of booleans if a haystack of array is passed
      */
     contains: (haystack, needle) => {
-		var type = $.type(haystack);
 		var ret = Mavo.Script.binaryOperation(haystack, needle, {
 			scalar: (haystack, needle) => {
+				var type = $.type(haystack);
 				if (type === "object") {
 					for (var property in haystack) {
 						ret = _.contains(haystack[property], needle);
@@ -344,15 +344,12 @@ var _ = Mavo.Functions = {
 						}
 					}
 				}
-				else if (type === "array") {
-					return _.contains(haystack, needle);
-				}
 				else {
 					return _.search(haystack, needle) >= 0;
 				}
+				return ret;
 			}, 
 		});
-
 		return ret;
     },
 
--- a/src/functions.js
+++ b/src/functions.js
@@ -325,7 +325,9 @@ var _ = Mavo.Functions = {
 	len: text => str(text).length,
 
 	/**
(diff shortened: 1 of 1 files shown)


src/functions.js
```diff
@@ -324,6 +324,36 @@ var _ = Mavo.Functions = {
...
+     * Search if a group, collection, or primitive contains needle
+	 * @returns Boolean if a haystack of object or primitive is passed
+	 * @returns Array of booleans if a haystack of array is passed
+     */
+    contains: (haystack, needle) => {
+		var ret = false;
+		if ($.type(haystack) === "object") {
+			for (var property in haystack) {
+				ret = _.contains(haystack[property], needle);
+				if ($.type(ret) === "array") {
```
LeaVerou: Prefer `Array.isArray()`, it's more readable and more efficient.

src/functions.js
```diff
@@ -324,6 +324,36 @@ var _ = Mavo.Functions = {
...
+				if (ret) { 
+					return true; 
+				}
+			}
+		}		
+		else if ($.type(haystack) === "array") {
+			ret = [];
+			for (var i = 0; i < haystack.length; i++) {
+				ret.push(_.contains(haystack[i], needle));
+			}
```
LeaVerou: This works well, but using `haystack.map()` would be more readable/elegant.

src/functions.js
```diff
@@ -324,6 +324,36 @@ var _ = Mavo.Functions = {
...
+    contains: (haystack, needle) => {
+		var ret = false;
+		if ($.type(haystack) === "object") {
+			for (var property in haystack) {
+				ret = _.contains(haystack[property], needle);
+				if ($.type(ret) === "array") {
+					ret = Mavo.Functions.or(ret);
+				}
+				if (ret) { 
+					return true;
```
LeaVerou: Why this? Won't the `return ret` in the end suffice?

LeaVerou:
Note that it still doesn't handle the case where `needle` is an array.
I'd repeat my recommendation for looking into `Mavo.Script.binaryOperation()`, which would handle both that case and the array case (that you are currently handling manually).

src/functions.js
efeichen: If ret is false it shouldn't be returned and terminate the function yet since there might be a match later in the object.

efeichen:
@LeaVerou If `needle` is an array, how should the haystack be searched? If a haystack of object is passed should it return true if the object contains all items in the array? Or does the object has to have a nested array of the exact same content as the needle array?

src/functions.js
```diff
@@ -324,6 +324,38 @@ var _ = Mavo.Functions = {
...
+						ret = _.contains(haystack[property], needle);
+						if (Array.isArray(ret)) {
+							ret = Mavo.Functions.or(ret);
+						}
+						if (ret) { 
+							return true;
+						}
+					}
+				}
+				else if (type === "array") {
```
LeaVerou: Have you tested without this else if? `binaryOperation()` should handle this case...

src/functions.js
```diff
@@ -324,6 +324,38 @@ var _ = Mavo.Functions = {
 
 	len: text => str(text).length,
 
+	/**
+     * Search if a group, collection, or primitive contains needle
+	 * @returns Boolean if a haystack of object or primitive is passed
+	 * @returns Array of booleans if a haystack of array is passed
+     */
+    contains: (haystack, needle) => {
+		var type = $.type(haystack);
```
LeaVerou: This should be inside the scalar function.

src/functions.js
efeichen: Yes! I deleted that block and nothing broke.

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+	/**
+     * Search if a group, collection, or primitive contains needle
+	 * @returns Boolean if a haystack of object or primitive is passed
+	 * @returns Array of booleans if a haystack of array is passed
+     */
+    contains: (haystack, needle) => {
+		var ret = Mavo.Script.binaryOperation(haystack, needle, {
+			scalar: (haystack, needle) => {
+				var type = $.type(haystack);
+				if (type === "object") {
```
LeaVerou: We don’t need the `type` variable here anymore, we're only using this once. The whole point of the variable was to cache the results so we don't have to run the function twice.

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+					for (var property in haystack) {
```
LeaVerou: We don’t need the type variable here anymore, we're only using this once. The whole point of the variable was to cache the results so we don't have to run the function twice.

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+				}
+				else {
+					return _.search(haystack, needle) >= 0;
+				}
+				return ret;
+			},
+			identity: null
+		});
+		// if result is an empty array, return false
+		if (ret.length === 0) {
```
LeaVerou: Very minor nitpick: Blank lines can often improve readability (provided we don’t overuse them!). My usual rule of thumb is I use blank lines to separate logical blocks of code, or control flow statements, unless they are the only ones within their parent block, in which case indentation suffices. You can look at the rest of the code for examples.

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+    },
+
 	/**
 	 * Case insensitive search
 	 */
-	search: (haystack, needle) => haystack && needle? str(haystack).toLowerCase().indexOf((needle + "").toLowerCase()) : -1,
+	search: (haystack, needle) => {
+		haystack = str(haystack);
+		needle = str(needle);
+		return haystack && needle? haystack.toLowerCase().indexOf((needle).toLowerCase()) : -1;
```
LeaVerou: You don't need the parens around needle here.

LeaVerou:
Huh, I thought I had made these comments before but I couldn't find them. Given that you just addressed them before I even posted them, I assume I had actually posted them before. Are there two open PRs about the same changes?!

efeichen:
Your comment on `type` came in a few minutes ago, which made me remember that I only revised contains() in the other PR (the one with lots of discussions on booleans) and not this, so I changed all of them before your comments came.

And yes, PR #358 is branched off of this branch, so that's why that PR has so many commits in the beginning I think

efeichen:
No, remember you told me to make a separate PR to change the binaryOperation() function in mavoscript.js? That's what #358 is for. The thing is I think it contains changes made by this PR's branch too because I branched that PR's branch off of this one.

Anyway, if everything looks good, I think merging the #358 will also add contains() function to functions.js, so I will close this one

Issue #177, opened 2017-05-14
Multiple mv-aliases


Dsan10s:
- Allow through space separation

i.e.

<... mv-alias="alias1 alias2">

PR #355, opened 2018-06-19
Multiple mv-alias

jwass91:
Addresses #177.

### DIFF
--- a/src/group.js
+++ b/src/group.js
@@ -228,12 +228,24 @@ var _ = Mavo.Group = $.Class({
 
 		var copy; // to handle renaming
 
+		var currentAlias; // store which alias we are copying from
+
 		this.propagate(obj => {
 			var propertyData = data[obj.property];
 
-			if (obj.alias && data[obj.alias] !== undefined) {
-				copy = copy || $.extend({}, data);
-				propertyData = data[obj.alias];
+			// find first alias with data, load that data, and set to be copied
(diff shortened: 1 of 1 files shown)


jwass91:
Tests for this addition are [here](https://github.com/mavoweb/test/pull/6).

src/group.js
```diff
@@ -231,9 +231,20 @@ var _ = Mavo.Group = $.Class({
 		this.propagate(obj => {
 			var propertyData = data[obj.property];
 
-			if (obj.alias && data[obj.alias] !== undefined) {
-				copy = copy || $.extend({}, data);
-				propertyData = data[obj.alias];
+			// find first alias with data, load that data, and set to be copied
+			if (obj.alias) {
+				aliasesArr = obj.alias.split(" ");
```
LeaVerou: I'd be inclined to split where `this.alias` is set so that we don't have to do it on every render. But this is fine too!

src/group.js
```diff
@@ -231,9 +231,20 @@ var _ = Mavo.Group = $.Class({
...
+				var i = 0;
+				while (i<aliasesArr.length){
```
LeaVerou: Is there a reason this while is not a for loop? It looks like a textbook for loop converted to use while…

Also, please install ESLint (and an ESLint plugin for your editor) which will help you follow the code style of the project. :)

PR #356, opened 2018-06-19
first/last non null element(s)

jwass91:


jwass91:
Tests for this addition are [here](https://github.com/mavoweb/test/pull/7).

src/functions.js
```diff
@@ -121,7 +120,30 @@ var _ = Mavo.Functions = {
 			return n && n > 1? [arr] : arr;
 		}
 
-		return n && n > 1? arr.slice(0, n) : arr[0];
+		if (n && n > 1) {
+			tempArr = [];
```
LeaVerou: You forgot var/let here which creates a global. Be **very** careful of that, it can be the source of super weird bugs.

Also, nitpick: For these kinds of variables (building a return value gradually), we typically use `ret` as a convention.

src/functions.js
```diff
@@ -121,7 +120,30 @@ var _ = Mavo.Functions = {
...
+			var i = 0;
+
+			while (i<arr.length && tempArr.length<n) {
```
LeaVerou: This would be more readable as a for loop.

src/functions.js
```diff
@@ -121,7 +120,30 @@ var _ = Mavo.Functions = {
...
+
+				if (arr[i]) {
```
LeaVerou: This will work with actual `null`, but not with the `null` we use in expressions, which is an object, and truthy just like every object (fun fact: even `new Boolean(false)` is truthy in JS).
Also, it will not add `0` and empty strings to the array, which you **do** want.
I'd suggest `Mavo.value(arr[i]) !== null`.

src/functions.js
```diff
@@ -133,7 +154,29 @@ var _ = Mavo.Functions = {
...
+
+			}
+
+			return ret;
+		} 
+		else {
+			var i = arr.length - 1;
+
+			while (i>=0 && Mavo.value(arr[i]) == null) {
+				i--;
```
LeaVerou: This can also be turned into a for loop. It's a little less typical than the first one, but still a for loop would be more readable. If you disagree, we can discuss in person.

src/functions.js
```diff
@@ -121,7 +120,27 @@ var _ = Mavo.Functions = {
 			return n && n > 1? [arr] : arr;
 		}
 
-		return n && n > 1? arr.slice(0, n) : arr[0];
+		if (n && n > 1) {
```
LeaVerou: Sorry, I didn't see this before. `n > 1` suffices here, no need for `n` as well. If `n` is greater than 1, it's definitely a positive number. (null etc all would return false).

One case to think about: Should we return an empty array or `null` if n is 0?

src/functions.js
```diff
@@ -121,7 +120,27 @@ var _ = Mavo.Functions = {
...
+					ret.push(arr[i]);
+				}
+
+			}
+
+			return ret;
+		} 
+		else {
+
+			for (i = 0; i<arr.length && Mavo.value(arr[i]) == null; i++) {
```
LeaVerou: Avoid the `==` operator unless you have a good reason to use it. Use `===` instead.
There is a lot of weirdness around it due to the way it does type coercion, and is the root of most JSWTFs. 
This is a fantastic article about type coercion: https://medium.freecodecamp.org/js-type-coercion-explained-27ba3d9a2839

src/functions.js
```diff
@@ -121,7 +120,27 @@ var _ = Mavo.Functions = {
...
-		return n && n > 1? arr.slice(0, n) : arr[0];
+		if (n && n > 1) {
+			var ret = [];
+
+			for (i = 0; i<arr.length && ret.length<n; i++) {
+
+				if (Mavo.value(arr[i]) !== null) {
+					ret.push(arr[i]);
+				}
+
```
LeaVerou: Nitpick: I generally agree that blank lines before/after control flow statements and blocks of code help readability, props for that! However, this here may be a little excessive.
I usually avoid blank lines if the block does not have any statements before/after it within the same scope since I feel that indentation sets it apart sufficiently, but no strong opinions here.

src/functions.js
```diff
@@ -121,7 +120,26 @@ var _ = Mavo.Functions = {
...
+			}
+
+			return ret;
+		} 
+		else if (n < 0) {
+			return _.last(Math.abs(n), arr);
+		}
+		else {
+			for (var i = 0; i<arr.length && Mavo.value(arr[i]) === null; i++) {}
+			return arr[i];
```
LeaVerou: A comment would be useful here, otherwise the reader has to read the whole code to understand that the else case means n is undefined.

Also why handle this separately and reimplement the searching algorithm? Isn't this basically the same as `first(1, arr)` except we return `ret[0]` (or null if ret is empty)?

src/functions.js
```diff
@@ -110,18 +110,35 @@ var _ = Mavo.Functions = {
...
-	// TODO return first/last non-null?
 	first: (n, arr) => {
 		if (arr === undefined) {
 			arr = n;
 			n = undefined;
 		}
 
 		if (!Array.isArray(arr)) {
-			return n && n > 1? [arr] : arr;
+			return n ? [arr] : arr;
```
LeaVerou: This will return `arr` if `n` is `undefined`. Change to `n !== undefined`.

src/functions.js
```diff
@@ -110,18 +110,35 @@ var _ = Mavo.Functions = {
...
+				}
+			}
+
+			return ret;
+		} 
+		else if (n < 0) {
+			return _.last(Math.abs(n), arr);
+		}
+		else { // n is undefined
+			return _.first(1, arr)[0];
```
LeaVerou: This will return `undefined` if the array has no non-null elements.
I didn't necessarily mean recursing btw, doing it in the main part of the code would be fine. 
You can get rid of negatives early, then define a variable that is `n === undefined? 1 : n` and run the loop based on that instead of n, then check if `n === undefined` at the end and return the first element if there is one or `null` otherwise.

PR #358, opened 2018-06-22
Enable Sliced Items To Operate With Identity 

efeichen:
Allow sliced array items in either argument `a` or `b` to be processed with identity, instead of simply copying them into the result array.

All past binary operations work the same, except `and([0, 1, 2, 3], [true, false])` which now returns `[false,false,true,true]`


karger:
per our discussion yesterday, i would say that (True and 2) should return 2, not True.  True should be treated as an identity for and, meaning that it returns the other value unchanged.

efeichen:
@karger Ok, I guess there was a slight miscommunication there. But now my question is, does identity (or "default") always have to try to return the other value unchanged? For operations like `subtract()` doesn't the user expect the leftover array items of the uneven array to either subtract or be subtracted *by* 0?

efeichen:
Also in JS `2 && true` returns `true`, so if we want to leave that unchanged here, true might not actually be the right identity for this?

efeichen:
Ok, I think this problem can be solved when I implement the right and left identity and scalar. This PR is to support the `contains()` function so it does not return leftover items of the longer array without doing operation with the identity

karger:
well that's certainly an interesting problem.  I'd say JS is just wrong here, so the question is whether we want to be consistent or right.  

subtraction doesn't have a (two sided) identity. 0 is a right identity (and leaves the left arg unchanged) but there is no left identity.

efeichen:
> well that's certainly an interesting problem. I'd say JS is just wrong here, so the question is whether we want to be consistent or right.

I feel like that depends on what Mavo users would want and() to return. If they want to use and() for their application, maybe chain it with another more complex Mavo function, I feel like it might be better to just have everything converted to booleans rather than numbers to avoid weird JS behaviors. But again, there might be cases where the numbers might come handy.

efeichen:
> subtraction doesn't have a (two-sided) identity. 0 is a right identity (and leaves the left arg unchanged) but there is no left identity.

I used a `leftScalar` to solve the same problem with `divide()`, since if the second arg is the longer one, you can never use an identity to return the leftover array item unchanged. So this could possibly be applied to subtraction as well.

I just did `leftScalar: (a, b) => b` for that one single leftover array item and it will just return that item unchanged. What do you think? Let me know if you need me to clarify anything I just typed

src/mavoscript.js
```diff
@@ -29,7 +29,7 @@ var _ = Mavo.Script = {
 			if (Array.isArray(a)) {
 				result = [
 					...b.map((n, i) => o.scalar(a[i] === undefined? o.identity : a[i], n)),
-					...a.slice(b.length)
+					...a.slice(b.length).map((n) => o.scalar(n, o.identity))
```
LeaVerou: I wonder if at this point it would be more readable to just loop over both, like so:

```js
var max = Math.max(a.length, b.length);
var ret = [];

for (let i=0; i<max; i++) {
	ret[i] = o.scalar(
		a[i] === undefined? o.identity : a[i],
		b[i] === undefined? o.identity : b[i]
	);
}

return ret;
```

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+    },
+
 	/**
 	 * Case insensitive search
 	 */
-	search: (haystack, needle) => haystack && needle? str(haystack).toLowerCase().indexOf((needle + "").toLowerCase()) : -1,
+	search: (haystack, needle) => {
+		haystack = str(haystack);
+		needle = str(needle);
+		return haystack && needle? haystack.toLowerCase().indexOf((needle).toLowerCase()) : -1;
```
LeaVerou: You don't need any parentheses around `needle`.

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+				}
+				return ret;
+			},
+			identity: null
+		});
+		// if result is an empty array, return false
+		if (ret.length === 0) {
+			return false;
+		}
+		return ret;
```
LeaVerou: Nitpick: Blank lines between blocks of code (especially before control flow statements if they are not the only statements inside their parent) can improve readability. Check the rest of the code for examples.

This is up to you though, it won't hold up the PR.

src/functions.js
```diff
@@ -324,10 +324,48 @@ var _ = Mavo.Functions = {
...
+	/**
+     * Search if a group, collection, or primitive contains needle
+	 * @returns Boolean if a haystack of object or primitive is passed
+	 * @returns Array of booleans if a haystack of array is passed
+     */
+    contains: (haystack, needle) => {
+		var ret = Mavo.Script.binaryOperation(haystack, needle, {
+			scalar: (haystack, needle) => {
+				var type = $.type(haystack);
+				if (type === "object") {
```
LeaVerou: We assigned `$.type()` to a variable to use it twice without invoking the function twice (i.e. for performance). Since the second clause has been removed, we don't need this anymore.

src/mavoscript.js
efeichen: Ya, considering my potential future implementations for left and right scalars and identities, I think this would be quicker to pick up.

LeaVerou: Probably faster too (though I imagine the difference is negligible).

LeaVerou:
> per our discussion yesterday, i would say that (True and 2) should return 2, not True. True should be treated as an identity for and, meaning that it returns the other value unchanged.
> well that's certainly an interesting problem. I'd say JS is just wrong here, so the question is whether we want to be consistent or right.
> Ok, I think this problem can be solved when I implement the right and left identity and scalar. This PR is to support the contains() function so it does not return leftover items of the longer array without doing operation with the identity.

I’m not sure I follow. In JS, `true && 2` **does** return `2`. Note that `2 && true` returns `true` however, i.e. `&&` is not commutative.

Since this is not an actual identity, but a default, I'm not sure if `true` is a good default. It seems to me that anything AND empty should return a falsy value.

> subtraction doesn't have a (two sided) identity. 0 is a right identity (and leaves the left arg unchanged) but there is no left identity.

0 is not an identity for both sides, but I think it's still a good **default**. And since we're renaming `identity` to `default`, I think it should stay.

joyously:
> per our discussion yesterday, i would say that (True and 2) should return 2, not True. True should be treated as an identity for and, meaning that it returns the other value unchanged.

For those users that are programmers, should it be clear that `and()` does not short-circuit?
I would expect `(True and 2)` to return `True`, not `2`.

src/mavoscript.js
efeichen: Yes. But for my future implementation of right and left scalars, I may need to use the original syntax since I could be using a different scalar for the leftover items of an uneven array.

LeaVerou:
@joyously I just updated my answer, I was thinking of `or`, not `and`. 😳

Although do keep in mind that neither and/or nor `if()` in Mavo do any control flow. Any expressions in both operands are evaluated fully before a value is returned. However, since expressions have no side effects, this should not matter for the output.

LeaVerou:
@joyously I think you may be thinking of OR, not AND. It's OR that short-circuits like that, i.e. `2 || true` does return `2` in JS. AND does not do that, since it needs to check all operands, so it returns the last one.

Do keep in mind that neither and/or nor `if()` in Mavo do any control flow. Any expressions in both operands are evaluated fully before a value is returned. However, since expressions have no side effects, this should not matter for the output.

joyously:
OK, but `(False and 2)` is `False`, and short-circuits in JS. Just playing devil's advocate...

LeaVerou:
`false and 2` should certainly return `false` in Mavo as well. No short-circuiting though.

src/mavoscript.js
```diff
@@ -23,14 +23,17 @@ var _ = Mavo.Script = {
 	},
 
 	binaryOperation: function(a, b, o = {}) {
-		var result;
+		var result = [];
```
LeaVerou: This only applies to the conditional below, so it's confusing to set it to an array here.

efeichen:
To summarize, in `and()`, all numbers that acts as booleans are converted to typical booleans with `!!num` which would always make `&&` operations commutative i.e. both `true && 2` and `2 && true` would return true. Therefore, when a number is operated with the current identity/default (`true`), Mavo will never return the argument back unchanged. I don't think neither @karger and @LeaVerou want to keep it that way, so let's come up with a consensus on that and I can fix it.

Personally, I agree that if we think identity as default, `false` makes more sense.

LeaVerou:
> All past binary operations work the same, except `and([0, 1, 2, 3], [true, false])` which now returns `[false,false,true,true]` 

Is this still true? I would expect `[0, false, 2, 3]`, since the operator is applied element-wise and apart from that is just a normal JS `&&`.

LeaVerou:
> To summarize, in and(), all numbers that acts as booleans are converted to typical booleans with !!num which would always make && operations commutative i.e. both true && 2 and 2 && true would return true. Therefore, when a number is operated with the current identity/default (true), Mavo will never return the argument back unchanged. I don't think neither @karger and @LeaVerou want to keep it that way, so let's come up with a consensus on that and I can fix it.

Just saw this.
No, we definitely don't want to convert them to booleans, and we don't need to. We don't need and to be commutative. I was stating that as a fact, not as a problem.

efeichen:
> Is this still true? I would expect [0, false, 2, 3], since the operator is applied element-wise and apart from that is just a normal JS &&.

This is still true, but I can easily make it `[0, false, 2, 3]` by getting rid of the double `!`s in `and` operator's scalar.

LeaVerou:
Merged! Please document the new function and send a PR if you haven't done so already.

Issue #336, opened 2018-05-16
Improve Github pull request flow


LeaVerou:
Currently, users without commit permissions can save data in their own fork and even send an "edit suggestion" (pull request). However, when they revisit the Mavo, they see the master data, not their own data. To see their own data, they need to use `?storage=...` in the URL, and I doubt anybody will have bookmarked that. This results in them making duplicate changes, as well as an overall frustrating user experience.

It would be easy to just check if a user has push permissions **when they log in**, not when they save, and load their own data if not, and if they have their own data copy.

This would also be a good first PR.

LeaVerou:
@jwass91 should this be closed? I believe you fixed that and there's now a message asking them if they want to switch, right?

jwass91:
@LeaVerou Yes, I believe this can be closed.

PR #360, opened 2018-06-22
GitHub backend improvements

jwass91:
Improves workflow of github backend (addresses #336)
- On login if a user has a fork of the current mavo: give an option to switch to that fork
- When a user is logged in and viewing their fork: give an option to send a PR or revoke a PR
- When a user saves edits on another mavo (edits are saved to their fork): update the URL to include `storage = their fork`

@karger I know this UI is different from what we were discussing earlier, but @LeaVerou and I thought best to get this first version up as a PR (before scraping it to began trying to implement an entirely different more complex UI) as it is still a major improvement from the current workflow. 

@karger Could you please outline the way you were saying to distinguish the two use cases.


src/ui.message.js
```diff
@@ -74,6 +82,10 @@ var _ = Mavo.UI.Message = $.Class({
 			$.remove(this.element);
 			this.closed.resolve(resolve);
 		});
+	},
+
+	nonAnimatedClose: function() {
+		$.remove(this.element);
```
LeaVerou: We don't want a second function just to prevent animation!
We already have the option stored, we can check it. Or we can just pass a duration of `0` to `$.transition()`. Not to mention that this function doesn't even resolve `this.closed`.

src/backend.github.js
```diff
@@ -253,13 +313,70 @@ var _ = Mavo.Backend.register($.Class({
...
 						return this.request(`repos/${this.username}/${this.repo}`)
-							.then(repoInfo => {
-								if (this.branch === undefined) {
-									this.branch = repoInfo.default_branch;
+						.then(repoInfo => {
+							if (this.branch === undefined) {
+								this.branch = repoInfo.default_branch;
+							}
+
+							let params = (new URL(document.location)).searchParams;
```
LeaVerou: No need to write `document.location`, `location` suffices.

src/backend.github.js
```diff
@@ -253,13 +313,70 @@ var _ = Mavo.Backend.register($.Class({
...
+
+							if (!this.mavo.source && !params.get("source")) { // if url doesn't have source, check for forks
```
LeaVerou: `source` is not the only URL param, `source-APPID` also works. But why do you need to check the URL param? If it's valid, there will be a `this.mavo.source` which you're already checking.

src/backend.github.js
```diff
@@ -461,7 +459,7 @@ var _ = Mavo.Backend.register($.Class({
 
 				history.pushState({}, "", `${location.pathname}?${params}`);
 				window.location.replace(`${location.pathname}?${params}`);
```
LeaVerou: Same about `window.location` here, you don’t need `window.`. In fact, you never need `window.`.

src/backend.github.js
```diff
@@ -191,20 +188,21 @@ var _ = Mavo.Backend.register($.Class({
 
 		if (this.notice) {
 			lastNoticeName = this.notice.o.name;
-			this.notice.nonAnimatedClose();
+			this.notice.o.closeTransition = false;
```
LeaVerou: I’m a little confused, I see you added a `style` parameter, why do you also need `closeTransition`?

src/ui.message.js
```diff
@@ -16,11 +16,8 @@ var _ = Mavo.UI.Message = $.Class({
 			[this.mavo.bar? "after" : "start"]: (this.mavo.bar || this.mavo).element
 		});
 
-		if (o.animationOff) {
-			this.element.style.WebkitTransition = "none";
-			this.element.style.WebkitAnimation = "none";
-			this.element.style.animation = "none";
-			this.element.style.transition = "none";
+		if (o.style) {
+			this.element.style = o.style;
```
LeaVerou: Please don't do that. It doesn't play nicely with other code, since it overwrites the entire inline style of the element. Better have the style param be a nicely readable object literal and use `$.style()` to apply it.

src/ui.message.js
```diff
@@ -78,15 +75,18 @@ var _ = Mavo.UI.Message = $.Class({
...
+		if (this.o.closeTransition) {
+			$.transition(this.element, {opacity: 0}).then(() => {
+				$.remove(this.element);
+				this.closed.resolve(resolve);
+			});
+		}
+		else {
 			$.remove(this.element);
 			this.closed.resolve(resolve);
-		});
```
LeaVerou: I’m not a huge fan of this, we are repeating the closing code twice. Right now it's only 2 lines, but in the future it could be more, and just like all WET code, it means that the repeated snippets can easily get out of sync by someone making changes hastily.
Did setting the duration to 0 not work?
In that case, I'd suggest using `Promise.resolve()` instead of `$.transition()` and then the `.then()` part is the same.
Also, you don't need `closeTransition` if I'm reading this right, just reading the computed style for a transitionDuration should work (but I may be missing something, I'm reviewing this in a rush at the airport right before boarding).

src/backend.github.js
```diff
@@ -324,6 +441,31 @@ var _ = Mavo.Backend.register($.Class({
...
+				dismiss: ["button", "submit"]
+			});
+
+			this.notice.closed.then(form => {
+				if (!form) {
+					return;
+				} 
+
+				history.pushState({}, "", `${location.pathname}?${params}`);
+				location.replace(`${location.pathname}?${params}`);
```
LeaVerou: Have you verified this doesn't get triggered if the notice is just dismissed?

src/ui.message.js
```diff
@@ -5,6 +5,7 @@ var _ = Mavo.UI.Message = $.Class({
 		this.mavo = mavo;
 		this.message = message;
 		this.closed = Mavo.promise();
+		this.o = o;
```
LeaVerou: `o` is convenient for a local variable, but means nothing as a property. The conventional name for this is `this.options`.

src/backend.github.js
jwass91: Yes

src/ui.message.js
jwass91: I switched that. But when making the change I noticed that there had been a previous bug in the code so I fixed that too. 

`this.options.style.transition === "none" ? Promise.resolve() : $.transition(this.element, {opacity: 0});` throws an error when this.options.style is undefined so I had to rewrite it to

```
if (this.options.style) {
	var promise = this.options.style.transition === "none" ? Promise.resolve() : $.transition(this.element, {opacity: 0});
} 
else {
	var promise = Promise.resolve();
}
``` 

I am pushing the changes now

src/ui.message.js
```diff
@@ -75,7 +75,12 @@ var _ = Mavo.UI.Message = $.Class({
 	},
 
 	close: function(resolve) {
-		var promise = this.o.style.transition === "none" ? Promise.resolve() : $.transition(this.element, {opacity: 0});
+		if (this.options.style) {
+			var promise = this.options.style.transition === "none" ? Promise.resolve() : $.transition(this.element, {opacity: 0});
+		} 
+		else {
+			var promise = Promise.resolve();
+		}
```
LeaVerou: Why the change from a ternary to a multi-line conditional? 

Also, `promise` is not exactly a descriptive variable name. It's like calling a number variable `number` or a text variable `string`. Perhaps `closed` or something?

src/ui.message.js
LeaVerou: You can use `$.value(this.options, "style", "transition") === "none"` to avoid such errors.

jwass91: I just rewrote it to `var closed = this.options.style && this.options.style.transition !== "none" ?  $.transition(this.element, {opacity: 0}) : Promise.resolve();`. Is that fine?

src/ui.message.js
```diff
@@ -75,13 +75,8 @@ var _ = Mavo.UI.Message = $.Class({
...
 	close: function(resolve) {
-		if (this.options.style) {
-			var promise = this.options.style.transition === "none" ? Promise.resolve() : $.transition(this.element, {opacity: 0});
-		} 
-		else {
-			var promise = Promise.resolve();
-		}
-		promise.then(() => {
+		var closed = this.options.style && this.options.style.transition !== "none" ?  $.transition(this.element, {opacity: 0}) : Promise.resolve();
+		closed.then(() => {
```
LeaVerou: If you set `transition` to `none`, `transitionDuration` becomes `0s` anyway, so you can simplify this a lot by just setting the duration to whatever the computed style of `transitionDuration` is (converted to a number of milliseconds).
Relevant: https://blissfuljs.com/docs.html#fn-transition

jwass91: I tried: 
```
$.transition(this.element, {opacity: 0, duration: this.element.style.transitionDuration})
.then(() => {
	$.remove(this.element);
	this.closed.resolve(resolve);
});
```
But it appears to animate even when duration is 0.

LeaVerou: Please note "(converted to a number of milliseconds)." in my message above. You specified a string like "0s" where a number was expected.
Also, you put the duration in the wrong place. Please look at the documentation I linked to above.
Also, better use the computed style, not the inline style, in case the user has specified something different in their CSS.

jwass91: The animation still shows up (flashes) with this: 
```
$.transition(this.element, {opacity: 0}, 0)
.then(() => {
	$.remove(this.element);
	this.closed.resolve(resolve);
});
```

LeaVerou: Aha! That's because there's a bug in Bliss, right here: https://github.com/LeaVerou/bliss/blob/master/bliss.shy.js#L513

Better fix the bug in its source than work around it. That's why we use our own libraries whenever possible :)

jwass91: Do you want me to fix the bug in Bliss?

LeaVerou: Yup!

FYI Bliss was originally a part of Mavo, but was released separately before it, to solicit more contributions from the community.

jwass91: I updated the Bliss to `(duration !== undefined && duration !== null ? duration : 400) + "ms"` but it still flashes. (I replaced the bliss.shy.min.js in my local Mavo directory with the new Bliss). Any ideas why it's not working?

joyously: Wouldn't you use 0 instead of 400?

LeaVerou: You don't need to take `null` into account. Catering for `undefined` is enough.

Or just `duration || duration === 0`

jwass91:
Just realized an issue with these changes. Now when a transition is not explicitly set there will be no effect. With the Mavo messages, transitions weren't explicitly set so the out transition was always 400. Now since I am calculating the transition duration, unless a transition is explicitly set, there will be no animation. So when you x out of a normal notification there isn't an animation. It might make sense to just go back to the old way where I handle the explicitly set option of transition = "none". Let me know your thoughts?

LeaVerou:
Hmmm. Maybe we should indeed check the inline style then, and default to 400ms if none is set.

LeaVerou:
Btw don't forget to update Bliss in this PR, now that your other PR is merged.

LeaVerou:
Can you please fix the merge conflicts?

Btw you don't need `.getPropertyValue("transition-duration")`, just `.transitionDuration` does the same thing.

jwass91:
I guess now I can switch it to `.transitionDuration` now that we check if there is a transition but when there isn't a transition (or transition duration) `.getPropertyValue("transition-duration")` will still return 0 whereas `.transitionDuration` returns undefined.

LeaVerou:
> when there isn't a transition (or transition duration) .getPropertyValue("transition-duration") will still return 0 whereas .transitionDuration returns undefined.

Simple test: open `about:blank` and run `getComputedStyle(document.body).transitionDuration`. You will see it returns "0s", which is the initial value for this property.

jwass91:
@LeaVerou Oops. I miss understood what you were saying. I thought you were saying `transitionDuration` on `style` (which didn't work). I will fix that and push.

LeaVerou:
Merged, thanks!

Btw, you almost never need `window.` (the main exception being when you want to reference a global and you have a local variable with the same name). `window.getComputedStyle` is the same as `getComputedStyle`.

PR #361, opened 2018-06-25
Implement prototypal right and left scalar and identity for binaryOperation()

efeichen:
As of now, the implementation is just a block of nested if statements checking if additional scalars or identities are implemented by the developer, but I'm looking for a less verbose way to do all this. Just hoping that my logic is going in the right direction though.

After that, I'm going to start changing all `identity` to `default` and also fix some weird behaviors with existing binary operations.


LeaVerou:
You definitely don't need this much code!
Just do something like:

```js
o.leftScalar = o.leftScalar || o.scalar;
o.rightScalar = o.rightScalar || o.scalar;
```

Then just use `o.leftScalar` and `o.rightScalar` everywhere.

src/mavoscript.js
```diff
@@ -27,10 +27,26 @@ var _ = Mavo.Script = {
...
+						...b.map((n, i) => o.scalar(a[i] === undefined ? o.rightDefault : a[i], n)),
+						...a.slice(b.length).map(n => o.rightScalar(n, o.rightDefault))
+					];
+				}
+				else {
+					result = [
+						...a.map((n, i) => o.scalar(n, b[i] === undefined ? o.leftDefault : b[i])),
+						...b.slice(a.length).map(n => o.leftScalar(o.leftDefault, n))
+					];
+				}
```
LeaVerou: This seems to be based on the older code. Please pull in changes and rebase.

src/mavoscript.js
```diff
@@ -64,7 +80,7 @@ var _ = Mavo.Script = {
 			});
 		}
 
-		o.identity = o.identity === undefined? 0 : o.identity;
+		o.default = o.default === undefined? 0 : o.default;
```
LeaVerou: I would expect a separate PR for the renaming. In general, please try to keep PRs small and atomic, it's much easier to review them which means they get merged faster.

LeaVerou:
I'm not sure whether this is ready for review, but it's still way too much code and way too much repetition. You don't need any of this branching logic. This is why we've used the max of the two lengths, so we don't need to compare them before looping.

efeichen:
Got it. I guess I'm still working on the habit of writing concise code. 😅

efeichen:
I shortened the for-loop by using chained ternary operators. If both array elements are NOT undefined, then it uses the regular scalar. If one of them is missing (undefined), then it uses left or right scalar, if defined, to return the desired result.

LeaVerou:
Why do you ever need to use `o.scalar()`? You've already assigned it to `o.leftScalar` and `o.rightScalar` as needed. No more checks are needed after that point, just use `o.leftScalar` and `o.rightScalar`!

LeaVerou:
@karger	could you take a quick look? It's only a few loc, and I want to make sure that's what you had in mind when you proposed `leftScalar` and `rightScalar`.

efeichen:
Also, I'm wondering if there are cases where an element in a Mavo list is undefined, so something like [1, undefined, 3]? Is that ever going to happen?

LeaVerou:
Not unless the user has done weird JS stuff. Mavo never produces that.

karger:
I'm going to nitpick two things:

1) you evaluate a[1]===undefined twice.  I would be inclined to set a local a variable once---it doesn't save LOC but I think it is clearer

2) at evaluation time you set o.leftScalar to o.scalar if it's undefined, and so forth.  I think this conflates two approaches in an inconsistent way.   Is the intention to allow the scalars and identities to be changed at runtime or not?  In the current approach, I can, at runtime, change o.leftScalar and have that change applied.  But if I change o.scalar, the change will *not* apply if I have already executed the function once, since the "cached" o.scalar in o.leftScalar will not be replaced.  So I recommend one of two consistent approaches:
    1) these things are set at compile time.  then I think it is more transparent to put the configuration of all these quantities in a separate "configure at load time" function that populates o.leftScalar from o.scalar etc.  this will let you assume they are all set when evaluated, which will make your runtime evaluation cleaner (no need to check for undefined values) and infinitesimally faster (for the same reason)
    2) these things are settable at runtime.  then you should not set any of the configuration fields like o.leftScalar, but instead collect the appropriate value into a local variable e.g.  leftScalar = o.leftScalar || o.scalar.  Then all changes made at runtime will apply, not only changes to some fields.

LeaVerou:
@karger thanks for reviewing. I mentioned 1 as well. 2 is a good point, let's assign to a local variable. Though I suspect there are very few cases where an operator is already there, but gets a leftScalar at a later point. Usually operators are added with their entire config.

I mostly wanted you to review this so you could verify that this is what you were suggesting for leftScalar and rightScalar, since I realized I didn't remember our discussion well enough. Basically, is there use in specifying all three and leftScalar/rightScalar are only used if a parameter is missing, whereas scalar is used when both are present? And in that case they don't need a leftIdentity/rightIdentity, right?

efeichen:
Might there be more complex cases where both a scalar and identity need to be set to produce the desired operation result?

karger:
I think the code is overgeneral/covering too many cases now.   Since the left scalar function is only called when the left argument is missing, the only value its left argument will ever have is the left identity.

You are combining two different approaches to the problem.  The simplest approach says "if the left argument is missing, here is the left identity value---pass it in to the binary function as a second argument, and similarly for the right argument missing."   The more general approach would say "if the left argument is missing, invoke the left scalar function on the *one* argument that is available."  You're doing both: defining a left identity, but then passing it in as an argument to the left scalar function.   This seems overkill.

The left/right identity approach is simpler when it works.  But it only works if an identity exists that will do the right thing in the binary function.  So for example 0 is a fine left (and right) identity for +, but if Infinity didn't exist in the language there would be no good identity for min, and you would instead want to use the leftscalar that returns its (only) argument unchanged.

In summary, I think leftscalar and rightscalar should only take a single argument.

Also note that for an awful lot of functions, the simple rule of "if there is only one argument, return that argument unchanged" is the right rule.  It might be nice to have a way to specify that without all the other complexity.

karger:
Additional note: what should the left scalar be for an operator like "<" ?  I would say that the natural generalization of "<" for other than 2 arguments is "all numbers in list are in increasing order" which would mean that if there is only a single argument the answer is always "true".  You could debate whether this still holds if the first argument is missing but the second exists, but I don't think there's anything simpler or more consistent.  

Is a null argument always the same as a missing argument?  Consider / .   If the second argument is missing you obviously return the first.  But what if the first is missing?  If you treat it as not there at all, you should by the same reasoning return the second argument unchanged.  But if you treat it as suggesting the default, maybe you could argue for returning 1/second argument.  I'd probably lean to always returning the (sole) argument unchanged.

efeichen:
> You are combining two different approaches to the problem.

@karger Should we allow the user of `binaryOperation()` to choose which approach they want to take/define then? I think that's what we came up with during the discussion couple weeks ago, if the scalar functions are not defined we would fallback to the identities.

> Is a null argument always the same as a missing argument?

In JS at least, if an argument is missing it's `undefined` not `null`, and they create different results. I prefer returning the argument value unchanged for `/` too, but I suspect that the Mavo users might expect something like `divide([1], [1, 2])` to return `[1, 0.5]`.

karger:
@efeichen that's an interesting question.  giving the programmer more options seems nice in general, but it does create the opportunity for inconsistency in what different programmers do.  arguably we might want to force consistency for future benefit.  But I lean towards keeping the flexibility.

You do raise an interesting point that "if the identities don't exist we fall back on the scalar functions" which means the identities have higher priority than the scalars---I'll just point out that is a choice (it could go the other way) and one I'm happy with.  In practice, something is wrong if the programmer is defining both an identity and a scalar, because only one can be used, but there's no nice way to stop them from doing it.

one other note: in the current code, if neither identity nor scalar is defined, you end up passing undefined as one argument to the underlying scalar function.  Which is not terrible since there is no way to recover from that error, but an alternative would be to raise an explicit error if a function with no identity or left/right scalar is called with only one argument.

efeichen:
> In practice, something is wrong if the programmer is defining both an identity and a scalar, because only one can be used, but there's no nice way to stop them from doing it.

By this, you mean defining identity and scalar on the same side right? I think it's fine for one to define a scalar on the left and identity on the right (e.g. the `minus` and `divide` functions). When someone does define both on the same side though, the identity would just be redundant because right now `binaryOperation` prioritizes right & left scalars.

> an alternative would be to raise an explicit error if a function with no identity or left/right scalar is called with only one argument.

There are functions, although not many, that still returns the desired results when no identities are declared. For example, I just found out that the `contains` function I implemented a few weeks ago still works without declaring `identity: null`, because `contains` calls `search` and `search(something, undefined)` would return `-1`, which is always the desired result for `contains` when only one argument is present. So for cases like this, it might be frustrating when an explicit error is still raised when everything works, but again, should we be lenient to the programmer or should we try to push them to stay consistant and always define an identity?

karger:
On 07/10/2018 05:21 PM, Kenneth Chen wrote:
>
>     In practice, something is wrong if the programmer is defining both
>     an identity and a scalar, because only one can be used, but
>     there's no nice way to stop them from doing it.
>
> By this, you mean defining identity and scalar on the same side right? 
> I think it's fine for one to define a scalar on the left and identity 
> on the right (e.g. the |minus| and |divide| functions). When someone 
> does define both on the same side though, the identity would just be 
> redundant because right now |binaryOperation| prioritizes right & left 
> scalars.
>
I wasn't talking about left vs right; just saying that it doesn't make 
sense  to have an identity and also a scalar on the same side, because 
if you have an identity the scalar will never be invoked.
>
>     an alternative would be to raise an explicit error if a function
>     with no identity or left/right scalar is called with only one
>     argument.
>
> There are functions, although not many, that still returns the desired 
> results when no identities are declared. For example, I just found out 
> that the |contains| function I implemented a few weeks ago still works 
> without declaring |identity: null|, because |contains| calls |search| 
> and |search(something, undefined)| would return |-1|, which is always 
> the desired result for |contains| when only one argument is present. 
> So for cases like this, it might be frustrating when an explicit error 
> is still raised when everything works, but again, should we be lenient 
> to the programmer or should we try to push them to stay consistant and 
> always define an identity?
>
yes, good question.  what does lea think?

karger:
I can imagine situations when leftScalar and rightScalar are identical, in which case it would be nice to just specify "scalar", but that name is already taken....

karger:
------ Original Message ------

efeichen:
@LeaVerou David and I talked about the problem with right and left scalar being the same and David came up with a new name for right and left scalar, **unary**. So when there's only one argument available, we would call `rightUnary` or `leftUnary` depending on which part is missing (same as before). When right and left unary are not different, however, the programmer can just define a `unary` and `binaryOperation` will just use that to evaluate both sides.

What do you think about this solution?

src/mavoscript.js
```diff
@@ -29,11 +29,19 @@ var _ = Mavo.Script = {
...
 				for (let i = 0; i < max; i++) {
-					result[i] = o.scalar(
-						a[i] === undefined ? o.identity : a[i],
-						b[i] === undefined ? o.identity : b[i]
-					);
+					if (a[i] !== undefined && b[i] !== undefined) {
+						result[i] = o.scalar(a[i], b[i]);
+					}
+					else if (a[i] === undefined) {
+						result[i] = o.leftScalar === undefined ? o.scalar(leftIdentity, b[i]) : o.leftScalar(b[i]);
```
LeaVerou: Since `o.leftScalar` is a function, just checking for truthiness works (i.e. `o.leftScalar ?` instead of `o.leftScalar === undefined ?`). For identities we explicitly compare with undefined, because 0 and "" are falsy.

LeaVerou: Since `o.leftScalar` is a function, just checking for truthiness works (i.e. `o.leftScalar ?` instead of `o.leftScalar === undefined ?`). For identities we explicitly compare with undefined, because 0 and "" are falsy.

src/mavoscript.js
```diff
@@ -29,11 +29,19 @@ var _ = Mavo.Script = {
...
 				var max = Math.max(a.length, b.length);
+				var leftIdentity = o.leftIdentity === undefined ? o.identity : o.leftIdentity;
+				var rightIdentity = o.rightIdentity === undefined ? o.identity : o.rightIdentity;
+
 				for (let i = 0; i < max; i++) {
-					result[i] = o.scalar(
-						a[i] === undefined ? o.identity : a[i],
-						b[i] === undefined ? o.identity : b[i]
-					);
+					if (a[i] !== undefined && b[i] !== undefined) {
```
LeaVerou: Nitpick: You are checking if `a[i]` is defined twice. You could do something like:

```js
if (a[i] === undefined) {
// ...
}
else if (b[i] === undefined) {
// ...
}
else {
// Both are defined if we got here
}
```

LeaVerou:
> @LeaVerou David and I talked about the problem with right and left scalar being the same and David came up with a new name for right and left scalar, **unary**. So when there's only one argument available, we would call `rightUnary` or `leftUnary` depending on which part is missing (same as before). When right and left unary are not different, however, the programmer can just define a `unary` and `binaryOperation` will just use that to evaluate both sides.

> What do you think about this solution?

Interesting idea, I'll think about it. I'm a little wary that we're adding a lot of complexity without having use cases for it. Did you discuss use cases?

efeichen:
> Did you discuss use cases?

No, we did not. David did say that he could think of situations when the two scalars are identical. I can ask him later.

LeaVerou:
I guess it can be as simple as `var leftUnary = o.leftUnary || o.unary;` or something along these lines, so I'm not opposed to add it.

src/mavoscript.js
```diff
@@ -29,11 +29,19 @@ var _ = Mavo.Script = {
...
+				var leftIdentity = o.leftIdentity === undefined ? o.identity : o.leftIdentity;
+				var rightIdentity = o.rightIdentity === undefined ? o.identity : o.rightIdentity;
+
 				for (let i = 0; i < max; i++) {
-					result[i] = o.scalar(
-						a[i] === undefined ? o.identity : a[i],
-						b[i] === undefined ? o.identity : b[i]
-					);
+					if (a[i] === undefined) {
+						result[i] = o.leftScalar ? o.scalar(leftIdentity, b[i]) : o.leftScalar(b[i]);
```
LeaVerou: The check is reversed, you are applying `o.leftScalar` when it's not there. Did you test this code?!

efeichen: My bad, thanks for checking.

LeaVerou:
Merged!

PR #362, opened 2018-06-26
Fix bug that skips second if-statement in ordinal function

efeichen:
Stumbled across a bug in `ordinal()` while documenting missing Mavo functions.


LeaVerou:
Whoa. Thanks!

Issue #359, opened 2018-06-22
SQL GROUP BY equivalent in Mavo expressions


LeaVerou:
We have identified several expressions (queries) that would benefit from a way to group items with the same value. 
For example, things like:

- Finding the average height of people by age
- Counting how many events in a month
- Templating items by e.g. tag

So far @jwass91, @karger and I have had several meetings about this and the resolutions so far are:
- There will be a `by` operator and corresponding `groupBy()` function (names TBB)
- They will return an array of objects with a `$value` and `$items` field, and a hidden flag to indicate that this is a grouped structure. `$value` will be the value that all items within the group share, `$items` will be an array of the items.
- It should be possible to group by multiple fields, potentially comma-separated, e.g. `person by age, height`. In that case, `$value` is an array of both values.
- `as` operator for naming fields. E.g. `person by floor(age / 10) as ageGroup` which would add an `ageCategory` field to the returned groups.
- If grouping by plain properties, they should also be added to the groups. E.g. `person by age` should return the same thing as `person by age as age`
- Aggregates detect grouped structures and return arrays of aggregates per group. E.g. `count(person by age)` returns an array of counts of people with the same age for each age. That also makes things like `(person by age) where age = max(age)` possible.
- Member expressions also detect grouped structures:
  - If the property is found on groups, they behave normally and return an array of these values. E.g. `(person by age).age` should return the same value as `unique(age)`, i.e. an array of all unique ages.
  - If the property is NOT found on groups, it passes through to `$items`. A grouped structure is returned, where the items for each group have been mapped to their values after the member operation. E.g. `(person by age).height` returns something like `[{$value: 
30, age: 30, $items: [160, 180, ...]}, ...]`

It is yet to be decided what the precedence should be for the new operator.

LeaVerou:
@karger when you read this, please visit the actual page because we've made a lot of edits!

PR #370, opened 2018-07-07
Initial groupBy

jwass91:
*Initial PR for #359*
- Adds `groupBy()` and `by` to group by one property
- Adds `as` to name fields 
- Updates aggregates to return arrays of value when passed "groupedBy" structures
- Tests for this PR [here](https://github.com/mavoweb/test/pull/9)

*Still need to implement:*
- Grouping by multiple properties / further grouping an already grouped structure

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -214,36 +214,91 @@ var _ = Mavo.Functions = {
 	 * Aggregate sum
 	 */
 	sum: function(array) {
-		return $u.numbers(array, arguments).reduce((prev, current) => {
-			return +prev + (+current || 0);
-		}, 0);
+		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+			var ret = [];
+
+			for (i in array) {
+				ret.push(_.sum(array[i]["$items"]));
+			}
+
+			return ret;
+		}
+		else {
+			return $u.numbers(array, arguments).reduce((prev, current) => {
+				return +prev + (+current || 0);
+			}, 0);
+		}
 	},
 
 	/**
 	 * Average of an array of numbers
 	 */
 	average: function(array) {
-		array = $u.numbers(array, arguments);
+		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+			var ret = [];
 
-		return array.length && _.sum(array) / array.length;
+			for (i in array) {
+				ret.push(_.average(array[i]["$items"]));
+			}
+
+			return ret;
+		}
+		else {
+			array = $u.numbers(array, arguments);
+
+			return array.length && _.sum(array) / array.length;
+		}
 	},
 
 	/**
 	 * Min of an array of numbers
 	 */
 	min: function(array) {
-		return Math.min(...$u.numbers(array, arguments));
+		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+			var ret = [];
+
+			for (i in array) {
+				ret.push(_.min(array[i]["$items"]));
+			}
+
+			return ret;
+		}
+		else {
+			return Math.min(...$u.numbers(array, arguments));
+		}
 	},
 
 	/**
 	 * Max of an array of numbers
 	 */
 	max: function(array) {
-		return Math.max(...$u.numbers(array, arguments));
+		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+			var ret = [];
+
+			for (i in array) {
+				ret.push(_.max(array[i]["$items"]));
+			}
+
+			return ret;
+		}
+		else {
+			return Math.max(...$u.numbers(array, arguments));
+		}
 	},
 
 	count: function(array) {
-		return Mavo.toArray(array).filter(a => !empty(a)).length;
+		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+			var ret = [];
+
+			for (i in array) {
+				ret.push(_.count(array[i]["$items"]));
+			}
+
+			return ret;
+		}
+		else {
+			return Mavo.toArray(array).filter(a => !empty(a)).length;
+		}
 	},
 
 	reverse: function(array) {
--- a/src/functions.js
+++ b/src/functions.js
@@ -214,7 +214,7 @@ var _ = Mavo.Functions = {
 	 * Aggregate sum
 	 */
 	sum: function(array) {
-		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+		if (array[0] && array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
 			var ret = [];
 
 			for (i in array) {
@@ -253,7 +253,7 @@ var _ = Mavo.Functions = {
 	 * Min of an array of numbers
 	 */
 	min: function(array) {
-		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+		if (array[0] &&array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
 			var ret = [];
 
 			for (i in array) {
@@ -271,7 +271,7 @@ var _ = Mavo.Functions = {
 	 * Max of an array of numbers
 	 */
 	max: function(array) {
-		if (array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
+		if (array[0] &&array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
 			var ret = [];
 
 			for (i in array) {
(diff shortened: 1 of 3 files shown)


src/functions.js
```diff
@@ -214,36 +214,90 @@ var _ = Mavo.Functions = {
...
+			return ret;
+		}
+		else {
+			return Math.max(...$u.numbers(array, arguments));
+		}
 	},
 
 	count: function(array) {
-		return Mavo.toArray(array).filter(a => !empty(a)).length;
+		if (array[0] && array[0]["$groupedBy"] === Mavo.groupedBy) { // grouped structures
```
LeaVerou: Remind me, what are we assuming this array is? The `$groupedBy` property (which should become `Mavo.groupedBy`, see comment below) is present on the entire grouped structure. So this is an array of grouped structures?! Why?

src/functions.js
```diff
@@ -214,36 +214,90 @@ var _ = Mavo.Functions = {
...
+			var ret = [];
+
+			for (i in array) {
```
LeaVerou: Avoid iterating over arrays with a `for..in` loop. They can often have other enumerable properties beyond their indices.
Prefer normal `for` loops, or `forEach()`.

src/functions.js
```diff
@@ -214,36 +214,90 @@ var _ = Mavo.Functions = {
...
+				ret.push(_.count(array[i]["$items"]));
+			}
+
+			return ret;
```
LeaVerou: There is a lot of repetition here for all 5 aggregates. We should probably leave them as-is and handle this separately, by wrapping each aggregate function with a function that checks if the argument is a grouped structure and calls the original function accordingly.

jwass91:
Sorry for the delay on this. 

I am happy to further discuss the implementation of `as` but this is the best I've come up with.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+						}
+						
+						obj["$items"] = items;
+						ret.push(obj);
+					}
+
+					return ret;
+				}
+				else { // invalid arguments, convert to arrays
+					// return (Mavo.toArray(array), Mavo.toArray(key));
```
LeaVerou: What happened here?
I'd just do this in the beginning of the function: `array = Array.isArray(array)? array : Mavo.toArray(array);` and same for key and then move on.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
 				return ret;
 			},
 			precedence: 3
-		}
+		},
+		"groupby": {
+			symbol: "by",
+			code: (array, key) => {
+				if (Array.isArray(array) && Array.isArray(key))  { 
+					var propName = key[Mavo.as] === true ? key[Mavo.name] : $.value(key[0], Mavo.toNode, "property");
```
LeaVerou: You don't need to do `=== true`, just checking `key[Mavo.as]` suffices.
Same for all the other Symbol properties.

Also, this is super minor, but can we call this variable `property` or `name`? (I don't think we have another one with that name in the same scope?). I generally try to use familiar variable names across scopes, as it gzips better (and aids readability if e.g. a property name is always called `property`). Also, there is recent research that abbreviations hurt code comprehension, so I avoid those too.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+			precedence: 2
+		},
+		"as": {
+			symbol: "as",
+			code: (property, name) => {
+				if (property !== undefined && $.type(property) === "array" && name !== undefined) {
+					if ($.type(name) === "string" || $.value(name[0], Mavo.toNode, "property") !== undefined) {
+						var ret = property.slice(0);
+						ret[Mavo.name] = $.type(name) === "string" ? name : $.value(name[0], Mavo.toNode, "property");
+						ret[Mavo.as] = true;
```
LeaVerou: But why have a separate `Mavo.name` property to hold the name? If `Mavo.name` is only present after an `as` operation, then its presence also signifies that it's an `as` expression. So, I would use only `Mavo.as` and have its value be the name.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+						return ret;
+					}
+
+					return property;
+				}
+				else if (property !== undefined) {
+					return property;
+				}
+				else {
+					return null;
```
LeaVerou: No need for `else` here, since all previous clauses return. Just `return null`, by itself, which makes it immediately obvious that it's a fallback.

In fact, you can just `return property` and leave it at that. I don't think there's much use in drilling down into what kind of falsy value exactly is `property`. One fallback for invalid data is sufficient.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+				}
+			},
+			precedence: 2
+		},
+		"as": {
+			symbol: "as",
+			code: (property, name) => {
+				if (property !== undefined && $.type(property) === "array" && name !== undefined) {
+					if ($.type(name) === "string" || $.value(name[0], Mavo.toNode, "property") !== undefined) {
+						var ret = property.slice(0);
```
LeaVerou: Super minor nitpick: `.slice()` works fine, no need for 0.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+					// return (Mavo.toArray(array), Mavo.toArray(key));
+				}
+			},
+			precedence: 2
+		},
+		"as": {
+			symbol: "as",
+			code: (property, name) => {
+				if (property !== undefined && $.type(property) === "array" && name !== undefined) {
+					if ($.type(name) === "string" || $.value(name[0], Mavo.toNode, "property") !== undefined) {
```
LeaVerou: You're checking if `name` is a string. What if `name` actually corrsponds to a property?
E.g. `foo as bar` and you already have a `bar` property?
IMO it should still work, and just extract the property from the value, via `Mavo.property`. Thoughts?

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+					var temp = new Mavo.BucketMap({arrays: true});
+					var ret = [];
+					ret[Mavo.groupedBy] = true;
+
+					for (var i = 0; i < array.length; i++) {
+						const k = i < key.length ? Mavo.value(key[i]) : null;
+						temp.set(k, Mavo.value(array[i]));
+					}	
+
+					for (var [value, items] of temp) {
```
LeaVerou: Avoid using `for..of` loops for now. They are lovely, but when transpiled by Babel they produce a bloated mess. See: http://babeljs.io/repl#?babili=false&browsers=last%204%20versions%2C%20Chrome%20%3E%3D%2041%2C%20IE%2011&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFAbghpAlhEwILASggbwFAQR4C-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=false&presets=stage-2%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2

Use `array.forEach` instead.

src/mavoscript.js
```diff
@@ -321,7 +321,62 @@ var _ = Mavo.Script = {
...
+						var obj = {"$value": value};
+
+						if (propName !== undefined) {
+							obj[propName] = value;
+						}
```
LeaVerou: You don’t need quotes for $, they are perfectly valid identifier chars in JS. Same applies to `obj["$items"]` below, it can just be `obj.$items`.

We could shorten the 4 lines above into 1 as:

```js
obj.$value = obj[propName || '$value'] = value;
```

or:

```js
var obj = {$value: value, [propName || '$value']: value};
```

I'm in two minds about whether it's worth it, given a small loss of readability. Up to you!

src/mavoscript.js
jwass91: Oops - I meant to fix this and guess I forgot. Fixing it now.

src/mavoscript.js
jwass91: I agree that it should work in some form but I see two different options. 
Say `bar` is a property set to `"hello"`
(1) `foo as bar` would set `Mavo.as = "bar"`
(2) `foo as bar` would set `Mavo.as = "hello"`

I am implementing (1) in this commit and think it is the way to go but wanted to put both options out there and get your feedback. The issue I have with (2) is that I think most likely when someone uses `foo as bar` they don't realize that they have a property `bar` or are intending to set the value to `"bar"`. It seems a little confusing to me if it goes and gets named the value of `bar` given that `bar` could be of any type.

LeaVerou: I agree it should be 1, but excellent observation!

LeaVerou:
Changes look good. I was wondering if instead of editing each aggregate and doing this thing with `aggregateCaller` we could just have a bit of code under their definitions that loops over each aggregate function and wraps it with the necessary code. On one hand this is more DRY and makes it easier to add new aggregates, on the other hand, it reduces how understandable the code is to a newcomer. What do you think? Is the tradeoff worth it or not?

jwass91:
I remember writing tests. I'm going to look for them now - not sure if I ever created the PR.

jwass91:
@LeaVerou [Here](https://github.com/mavoweb/test/pull/9) is the updated tests PR.

PR #372, opened 2018-07-18
Use searchParam.set to populate req.data

efeichen:
Now, instead of setting the timestamp on the variable`call`, if the request method is `GET`, all the `searchParams` are set on `req.data`. My question is, is it going to make any difference now `timestamp` is set in `req.data`?

Also, I noticed that the workaround I proposed (concatenating `"&"` with the data params) is only used when the method is GET and timestamp is only set when the method is GET. So shouldn't the workaround always work, since timestamp will always be there?


efeichen:
Ummmm git for some reason chose to push all the changes from my Google Drive backend branch to here... So I reverted all of that.

LeaVerou:
It's generally not a good idea to just revert commits you don't want in there, instead investigate why you got those commits in your PR in the first place. That way you won't have to deal with this problem every time you send a PR.
I suspect you branched off from your Google Drive branch, which is why all these changes were there. You should instead go to the master branch and branch off from there.
Assuming you're gonna do that and submit a new PR, I’m gonna go ahead and close this, and am looking forward to the new one :)

efeichen:
Ya for sure. Making a new one ASAP.

PR #373, opened 2018-07-19
Use searchParam.set to populate req.data

efeichen:
This is the **new PR** after recreating the branch from master. Below are my old comments:

Now, instead of setting the timestamp on the variable`call`, if the request method is `GET`, all the `searchParams` are set on `req.data`. My question is it going to make any difference now `timestamp` is set in `req.data`?

Also, I noticed that the workaround I proposed (concatenating `"&"` with the data params) is only used when the method is GET and timestamp is only set when the method is GET. So shouldn't the workaround always work?

### DIFF
--- a/src/backend.js
+++ b/src/backend.js
@@ -102,7 +102,7 @@ var _ = Mavo.Backend = $.Class({
 				for (const p in req.data) {
 					call.searchParams.set(p, req.data[p]);
 				}
-				req.data = "";
+				delete req.data;
 			}
 			else {
 				req.data = JSON.stringify(req.data);
--- a/src/backend.js
+++ b/src/backend.js
@@ -84,27 +84,31 @@ var _ = Mavo.Backend = $.Class({
 			"Content-Type": "application/json; charset=utf-8"
 		}, req.headers || {});
 
-		req.data = data;
-
 		if (this.isAuthenticated()) {
 			req.headers["Authorization"] = req.headers["Authorization"] || `Bearer ${this.accessToken}`;
 		}
 
+		req.data = data;
+		var dataParam;
+
+		// Prevent getting a cached response. Cache-control is often not allowed via CORS
+		if (req.method == "GET") {
+			dataParam = new URL("https://mavo.io");
+			dataParam.searchParams.set("timestamp", Date.now());
+		}
+
 		if ($.type(req.data) === "object") {
 			if (req.method == "GET") {
-				req.data = Object.keys(req.data).map(p => p + "=" + encodeURIComponent(req.data[p])).join("&");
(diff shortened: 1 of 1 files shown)


LeaVerou:
> Also, I noticed that the workaround I proposed (concatenating `"&"` with the data params) is only used when the method is GET and timestamp is only set when the method is GET. So shouldn't the workaround always work?

Yes, it would *happen* to work, but it's a hack. What if in the future we decide to stop adding the timestamp because e.g. we've found a better way to address the issue that the timestamp param is there for? Nobody would think to change the "&" and code would break.
In general, try to write code with its future evolution in mind, to the extent possible.

src/backend.js
```diff
@@ -106,6 +97,17 @@ var _ = Mavo.Backend = $.Class({
 			call.searchParams.set("timestamp", Date.now());
 		}
 
+		if ($.type(req.data) === "object") {
+			if (req.method == "GET") {
+				Object.keys(req.data).map(p => call.searchParams.set(p, req.data[p]));
```
LeaVerou: Nitpick: `map()` implies that you are mapping one array to another, and plan to use the result. If you're just looping over an array, use `forEach` which communicates your intent better.

Also, in this case you don’t need either, just a for...in loop would do:

```js
for (var p in req.data) {
	call.searchParams.set(p, req.data[p]);
}
```

(make sure to check this works, I haven't)

src/backend.js
```diff
@@ -106,6 +97,17 @@ var _ = Mavo.Backend = $.Class({
...
+				req.data = call.search;
+				call.search = "";
```
LeaVerou: Why is this step needed? Isn't the whole point of doing this to add it as a query string to the call URL?
If it's to avoid adding it twice, you could just clear `req.data`.

src/backend.js
```diff
@@ -106,6 +97,18 @@ var _ = Mavo.Backend = $.Class({
 			call.searchParams.set("timestamp", Date.now());
 		}
 
+		if ($.type(req.data) === "object") {
+			if (req.method == "GET") {
+				for (const p in req.data) {
+					call.searchParams.set(p, req.data[p]);
+				}
+				req.data = "";
```
LeaVerou: `undefined` caused issues?

In general, have you verified that the tests that were passing before this change still pass?

efeichen: All the tests still pass. I think it was caused by another issue, and `undefined` was probably a false cause.

LeaVerou: I'd just do `delete req.data` which emulates never setting it in the first place (since we've incorporated it to the URL, we don't need it anymore as a separate parameter)

LeaVerou:
Merged!

PR #381, opened 2018-09-09
Fix: Logic & Comparison Operator Properties

efeichen:
1. Removed `logical` flag for `and` and `or` operators.
2. Renamed `logical` flag to `comparison` as they are only defined in comparison operators. Modified `addBinaryOperator()` accordingly.
3. Removed `reduce` method from `or` and modified `addBinaryOperator()` accordingly.

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -92,10 +92,10 @@ var _ = Mavo.Script = {
 				operands = operands.map(val);
 			}
 
-			var prev = o.logical? o.identity : operands[0], result;
+			var prev = o.comparison ? o.identity : operands[0], result;
 
 			for (let i = 1; i < operands.length; i++) {
-				let a = o.logical? operands[i - 1] : prev;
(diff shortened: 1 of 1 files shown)


src/mavoscript.js
```diff
@@ -233,22 +230,19 @@ var _ = Mavo.Script = {
...
 			scalar: (a, b) => a != b,
 			symbol: ["!="],
 			identity: true
 		},
 		"and": {
-			logical: true,
-			scalar: (a, b) => !!a && !!b,
-			identity: true,
+			scalar: (a, b) => a && b,
+			identity: false,
```
LeaVerou: Why are we changing the identity of `and` to `false`? That doesn't seem to make sense or be related to this PR.
Everything else looks ok.

efeichen: Oh ya my bad, that's not related to this PR.

LeaVerou:
Merged, thanks!

PR #382, opened 2018-09-10
Convert identities to defaults

efeichen:
Change the name `identity` to `default` for `binaryOperation`. Probably last prepping step before implementing metadata for Mavo functions.


LeaVerou:
Merged, thanks!

PR #393, opened 2018-10-08
Improve handling of empty expressions

### LINKED ISSUE #392: Empty expression leads to confusing error message
The error Mavo prints to the console is something like this.

Firefox:
```
Oops! 😳 There is something wrong with the expression : expected expression, got ')'
```
Chrome:
```
Oops! ðŸ˜³ There is something wrong with the expression : Unexpected token )
```
(there should be an emoji there, not sure what's up with encoding)

The JavaScript that an empty expression compiles to is basically `return ();`, which is why the parse error is what it is, but this error is not very friendly. Perhaps we should special-case empty expressions to have an error like "There is an empty expression!" Or, I wonder if users might expect an empty expression to evaluate to an empty string or something. Of course, as a programmer I prefer to see the error, but I'm not in the target audience.

This is something I ran into while playing with the inspector and it may be hard to accidentally run into this in the wild.

betaveros:
Don't treat empty expressions or expressions with only spaces as
expressions. However, still parse them out so that they don't interfere
with the next expression. If we are forced to evaluate such an
expression (e.g. via mv-value), evaluate it to the empty string.

Fixes #392.

### DIFF
--- a/src/expression.js
+++ b/src/expression.js
@@ -102,7 +102,8 @@ _.Syntax = $.Class({
 
 			if (/\S/.test(match[1])) {
 				ret.push(new Mavo.Expression(match[1]));
-			} else {
+			}
+			else {
 				// If the matched expression is empty or consists only of
 				// whitespace, don't treat it as an expression.
 				ret.push(match[0]);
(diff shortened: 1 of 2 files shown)


LeaVerou:
The actual code looks great!
Please install ESLint so you can follow the project code style.
Once you fix those few and extremely minor issues, this is ready to be merged.

betaveros:
OK, I ran eslint. For what it's worth, it reported two warnings elsewhere in the codebase that I didn't try to fix in this PR.

PR #396, opened 2018-10-19
Remove mv-pending-edit every time editing actually starts

### LINKED ISSUE #395: (Empty) primitive and its (empty) editor are simultaneously displayed
Steps to reproduce:

- Edit a primitive string property and set it to the empty string, e.g. "Just a span" in https://test.mavo.io/editor.html
- Turn edit mode off and on again

Result: The placeholder text "(Span)" is shown twice; the first copy is just text while the second copy is editable.

betaveros:
The class is added every time pre-editing potentially starts, so it
should be removed every time we know editing actually starts. This is
consistent with the restoration of tabindex. Fixes #395.

### DIFF
--- a/src/primitive.js
+++ b/src/primitive.js
@@ -326,8 +326,6 @@ var _ = Mavo.Primitive = $.Class({
 			this.preEdit.then(evt => {
 				$.unbind(this.element, ".mavo:preedit");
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks, merged! Sorry for the delay, I got sick.

PR #402, opened 2018-11-25
Correct divide() and and() operations

efeichen:
Modified `divide()` and `and()` to meet expected test results.

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -145,6 +145,7 @@ var _ = Mavo.Script = {
 		},
 		"divide": {
 			scalar: (a, b) => a / b,
+			rightUnary: b => b,
 			default: 1,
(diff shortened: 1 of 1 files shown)


LeaVerou:
Could you please explain the reasoning for this change?

Also, any chance you could take a look at https://github.com/mavoweb/mavo/pull/374 ? It's been sitting there for a while waiting for changes, and eventually it's gonna start having conflicts as Mavo evolves.

efeichen:
For `divide`, the change makes sure that when the left argument is missing, the default is not used as fallback and instead return the right argument as the quotient, which is what we discussed before as the optimal result. Likewise, we discussed over the summer that the default argument for `and` should be `false`.

Sorry about #374, I will make changes ASAP.

PR #406, opened 2018-12-02
Add "rd" option to ordinal function

efeichen:
I noticed that `ordinal(3)` returns `th` instead of `rd`, so I added the option to the function.


LeaVerou:
Thanks!

PR #407, opened 2018-12-02
Binary Operation: Allow parameter of type Function as argument

efeichen:
This feature is to support #374. When wrapping a function with `binaryOperation()`, one can pass the function with metadata as `o`. `o.scalar` is then assigned from `o`, and the function's metadata such as `unary` and `default` will be automatically assigned too.


LeaVerou:
Thanks!

Issue #411, opened 2018-12-08
Memory leak bug in popup editor


hopeful2:
A bug in popup editor and [a bug in Bliss.unbind method](https://github.com/LeaVerou/bliss/issues/221) work together to cause rapid memory leaks in Mavo apps.

## Details of the bug

Everytime a popup editor is shown, a hideCallback handler is redefined for the same editor, and registered as the callback function of click/focus events. Even though it is the same code, browsers treat it as a different new function object. 

The listener is removed when a user clicks somewhere else to hide the popup editor. hideCallback is a named function, and attached to a class as method. The bug may not be a big deal, and would otherwise be difficult to notice.

The listener is registered using Bliss.bind method. Bliss.bind searches the existing records of registered listeners, and since it is a "new" function, Bliss.bind creates a new record and pushes to the data array under Bliss.listeners.

A bug in Bliss.unbind causes it to remove event listeners from the DOM without going through Bliss.listeners, if both event type and call back function are provided. Thus the data arrays under Bliss.listeners keep increasing. If there are a few popup editors and users click around, there will soon be huge arrays with useless records. 

To verify the bug, please go to https://mavo.io/docs/primer, pick a popup editor, for example clicking to edit the birthday of Adam Catlace, click somewhere else to hide the popup.  In browser console, examine the data array by running

$.listeners.get(document).click

The data array should be empty after hiding the editor and removing the listener. But because of the bugs, you will see an array with one record in the console log. Now click on birthday to show the editor, click somewhere else to hide the editor,  and repeat the process for a few times. Check the array again. It has grown by the number of times showing the editor. 

The listeners are registered on element "document", which is never cleared from memory for the lifecycle of the page. So the WeakMap Bliss.listeners is powerless here in terms of garbage collection. 

## the solution

hideCallback should be defined in class constructor to run once only.

I will submit a PR to address this issue. I have submitted a PR to Bliss repo to address the bug there. 

best regards.

Oliver

LeaVerou:
Hey, thank you so much for all your research! I've replied to your individual PRs, looking forward to getting the fixes merged!

PR #412, opened 2018-12-08
fix memory leak bug in popup editor [#411]

hopeful2:
move the definition of hideCallback to constructor and define the function only once for the lifecycle of the popup editor.

### DIFF
--- a/src/ui.popup.js
+++ b/src/ui.popup.js
@@ -67,18 +67,18 @@ var _ = Mavo.UI.Popup = $.Class({
 		if (this.editor.matches("select")) {
 			this.editor.size = Math.min(10, this.editor.children.length);
 		}
+		var _popup = this;
+		this.hideCallback = evt => {
+			if (!_popup.element.contains(evt.target) && !_popup.primitive.element.contains(evt.target)) {
+				_popup.hide();
+			}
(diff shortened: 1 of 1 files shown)


src/ui.popup.js
```diff
@@ -67,19 +67,19 @@ var _ = Mavo.UI.Popup = $.Class({
 		if (this.editor.matches("select")) {
 			this.editor.size = Math.min(10, this.editor.children.length);
 		}
+		var _popup = this;
+		this.hideCallback = evt => {
+			if (!_popup.element.contains(evt.target) && !_popup.primitive.element.contains(evt.target)) {
+				_popup.hide();
+			}
+		};
```
LeaVerou: Hey, thanks!
You don't need `_popup`, arrow functions have no `this` binding, so they inherit that of the containing scope. So you can just use `this` in it, just like it was using below.

hopeful2: Thank you for being so open minded to PR.

I have changed it back to arrow function and verified the code works on my local dev server. I probably had a typo somewhere else and the dev console attributed the error to the arrow function.

LeaVerou:
Merged, thanks!

PR #414, opened 2018-12-09
track observer instance in mavo mv-load event handler fixes #413

### LINKED ISSUE #413: keep track of observer instance in Mavo constructor
I am trying a way to create and destroy Mavo instances dynamically and manage the memory at the same time. This will open up new opportunities to use Mavo in different ways.

To achieve this, we need to keep track of ALL objects, observers and event listeners,  and remove them completely when done. The current code base already keeps track most of these, with few exceptions.

One of the exceptions is the observer instance in the event handler of 'mv-load.mavo'.  Can this instance be attached as a member of Mavo instances? Are there any concerns or side effects? 

I will submit a PR shortly for your consideration and easy reference to the code in question.

hopeful2:
track this observer instance to remove it manually when necessary. otherwise the observer instance will stay in memory, keep reference to mavo and node instances, and prevent destroyed nodes from garbage collection.

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -223,9 +223,9 @@ var _ = self.Mavo = $.Class({
 							});
 						}
 
-						if (observer) {
-							observer.destroy();
-							observer = null;
+						if (this.observer) {
+							this.observer.destroy();
+							this.observer = null;
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


hopeful2:
Great. I wasn't sure whether this observer was left alone for a reason. 

I have renamed it as suggested.

Issue #404, opened 2018-11-29
Comparing lists of mismatched lengths extends with `true` (approximately, 1)


betaveros:
It is convenient that many Mavo binary operations naturally vectorize across lists so that you can add two lists and corresponding components will be added together, and the shorter list is padded to the length of the longer. However this breaks down when you compare two lists since comparison doesn't have a really sensible identity. It looks like the comparison operators are defined to have `true` as their identity, but this is the identity of `&&`, which is used in the interpretation of chaining comparisons. However this `true` identity is still used to pad lists of different lengths when comparing them, makes comparing two lists of different lengths somewhat nonsensical: e.g. `[1, 2, 3] < [4]` approximately becomes `[1 < 4, 2 < true, 3 < true]`.

I'm not really sure what the best behavior would be in this case. In a vacuum I would expect vectorized operators between lists of different lengths to either fail or use the length of the shorter operand, but it doesn't look like this is consistent with the goals or other parts of Mavo.

LeaVerou:
Oops, it definitely shouldn't be true. This is a good example of a case where we need separate defaults for left and right operands, then we just use `±Infinity` accordingly. Thoughts?

Issue #403, opened 2018-11-29
Chained comparison operators do not vectorize correctly


betaveros:
Assuming the following implicit collection:

```html
<div property="a">5</div>
<div property="a">5</div>
<div property="a">5</div>
<div property="b">3</div>
<div property="b">3</div>
<div property="b">3</div>
```

Mavo expressions `[a <= b]` and `[b <= a]` work as expected, resulting in a list of 3 falses and 3 trues respectively. One might expect `[a <= b <= a]` to chain and evaluate a vectorized `&&` across these two results, but it does not; I believe the underlying JavaScript evaluates `[false, false, false] && [true, true, true]`, which just results in the second operand `[true, true, true]` since JavaScript lists are truthy.

Somewhat entangled with #347 but mostly orthogonal. I think it is not hard to fix, though.

LeaVerou:
Wow, nice catch!

Some things I noticed:

The operators do combine correctly indeed (you can verify by running `Mavo.Script.compile("a <= b <= a")`), so I don't think it's particularly related to #347. On the other hand, running `Mavo.Functions.lte([3, 3, 3], [5, 5, 5], [3, 3, 3])` does correctly return `[false, false, false]`! Odd.

betaveros:
Actually I think #450 broke this again, so I should fix it again and add some tests.

PR #418, opened 2018-12-10
Rewrite comparisons with custon compare function

### LINKED ISSUE #347: Math comparison operators don't combine if different
I.e. while `1 < 2 < 3` is correctly interpreted as `1 < 2 and 2 < 3`, this does not happen with `-3 <= -3 < 1` which returns false, since it's interpreted as `(-3 <= -3) < 1 => true < 1 => 1 < 1 = false`

betaveros:
Define a new Mavo `compare` function that takes operands alternating
with names of comparison operators, so that mixed chained comparisons
work without requiring intermediate operands to be re-evaluated. Rewrite
comparisons to call this function. Also correctly handle vectorization
of comparisons in this function.

Fixes #347, #403, and #404.

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -118,7 +118,7 @@ var _ = Mavo.Script = {
 	},
 
 	/**
-	 * Mapping of operator symbols to function name.
+	 * Mapping of operator symbols (strings) to function names (strings).
 	 * Populated via addOperator() and addLogicalOperator()
 	 */
 	symbols: {},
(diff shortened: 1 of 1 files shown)


LeaVerou:
Merged, thanks! Could you please add a few tests too?

LeaVerou:
@betaveros Could you please fix the existing tests so that this change doesn't break them? http://test.mavo.io/mavoscript.html
Currently a bunch of tests are failing because of this change.

PR #421, opened 2018-12-10
clear timeout callback upon dismiss message fixes #420

### LINKED ISSUE #420: Proposal - clear dismiss timeout when closing message
While working on the memory issue, I have long been puzzled by the fact that Mavo instance and child node instances still stayed in the memory with detached elements, after removing ALL listeners and observers, and destroying all nodes. After days of monitoring the memory allocation and tracing, I have finally nailed down the last surviving reference holder - the callback of dismiss timeout function. 

Strictly speaking, this is not a bug and has little impact on app performance. I am filing an issue here and proposing changes to make future debugging easier. I will submit a PR shortly. Hopefully it will stop other developers from pulling their hair debugging similar issues. 

## details of the surviving reference holder

When a collection item is deleted, a small popup message confirms the item has been deleted. It offers a chance to undo, and a button to dismiss the message. Upon mouseleave event on the message, a timeout callback is registered, with context containing reference to "this" message. "This" message contains reference to the mavo instance, and the whole family of nodes/elements. The dismiss callback has a 20-seconds timeout, and thus is still in the memory at the end of recording the memory allocation timeline. This is the source of confusion and headaches. When this last reference is removed, the created objects and elements finally disappear from the memory.

Here is what happens at the end of timeout, with pseudo code illustrating the function invocation process. "message" is the instance of UI.Message class, and messageElement is the message box visible to users.

message.close() 
=> Bliss.transition()
=> Bliss.once(messageElement, ...)
=> Bliss.bind(messageElement, ... )
=> Bliss.unbind(messageElement, ... )
=> Bliss.remove(messageElement)
=> message.closed.resolve()

Note Bliss.bind always creates a record under Bliss.listeners for housekeeping. The record uses messageElement as key and contains a map of event handlers. This explains why an entry is added back mysteriously after it is removed from Bliss.listeners registry, approximately 20 seconds later.

Bliss.unbind will remove event listener and the internal housekeeping record, however it will not remove the key messageElement. Bliss.listeners is a WeakMap object and will remove this key automatically if there is no other reference to this element. On older browsers not supporting WeakMap, Bliss.listeners defaults to Map and this element key needs to be manually deleted. 

## The proposal

Track and clear timeout callback when the message dialog is closed by clicking the dismiss button. The callback context is garbage collected immediately. This will remove the reference to Mavo instance while recording the memory allocation timeline, so we can focus on other remaining retainers if any.

hopeful2:
The context of timeout callback contains reference to the message object and Mavo instance. The function is invoked 20 seconds later, and is easily overlooked when recording memory allocation timeline to debug memory related issues. The details of the rationale behind this PR is documented in issue #420. Please consider.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -263,7 +263,7 @@ var _ = Mavo.Functions = {
 		}
 
 		if (num < 10 || num > 20) {
-			var ord = ["th", "st", "nd", "th"][num % 10];
+			var ord = ["th", "st", "nd", "rd", "th"][num % 10];
 		}
 
 		return ord || "th";
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -23,6 +23,7 @@ var _ = Mavo.Script = {
 	},
 
 	binaryOperation: function(a, b, o = {}) {
+		o.scalar = typeof o === "function" ? o : o.scalar;
 		var result;
 
 		if (Array.isArray(b)) {
--- a/src/primitive.js
+++ b/src/primitive.js
@@ -461,7 +461,7 @@ var _ = Mavo.Primitive = $.Class({
 						// FIXME Once this is resolved with mousedown, every time we edit, evt is still mousedown regardless
 						// so this ends up focusing even when it shouldn't
 						if (evt && evt.type == "mousedown" || document.activeElement === this.element) {
-							this.editor.focus();
+							requestAnimationFrame(() => this.editor.focus());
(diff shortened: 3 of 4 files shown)


LeaVerou:
Thank you so much for all the research you're doing, we're so grateful!
This PR has the same issue as the previous one: Changes that have already been merged are in the diff. Could you please look into it?

Issue #410, opened 2018-12-05
Feature request - provide a method to destroy Mavo instance and release memory


hopeful2:
Hi,

Thank you for this awesome library. 

We are working on a project and use Mavo as editor. There will be multiple sections on the same page, each section is a Mavo app. Users can copy and add sections, and remove unwanted sections. There are potentially many sections on the page, and users activate a Mavo app when they want to edit the section. Users work on one section at a time and each section/app is independent. Ideally we'd like for users to toggle off/destroy the Mavo app when done, in order to avoid running too many apps on the same page. 

We have modified a local copy of Mavo.js and set defined Mavo property (Mavo.all[1], Mavo.all[2] etc) to be configurable so as to remove the instance later.  We have called the destroy method of observers, and used Bliss.unbind to remove event listeners from elements. 

However we have noticed the memory usage continues to increase after activating and destroying the Mavo instances. The issue exists only when Mavo is activated and destroyed.  Some event listeners/observers probably stay in the memory when a Mavo instance is destroyed. This will eventually lead to a forced garbage collection, or crash.

We have just found out that Bliss.unbind does not remove all event listeners when calling this function without event types or callbacks. It leaves some listeners behind if the listeners have been registered (Bliss.bind) by setting capture option to true.   I have submitted a PR to Bliss repo for you to consider and address this issue. 

Assuming all event listeners can be completely removed, what else shall we do to clean up? Your guidance is highly appreciated. Could you point to the right direction on this issue? Is it possible to completely remove Mavo instance and release memory? 

Update: we are using the latest development version (1.7beta) for tests. We have modified a local copy and replaced addEventListener with Bliss.bind, and everything else is the same. Now all event listeners are removed after destroying Mavo instance. However the memory issue still exists, and the number of dom nodes does not decrease.

According to Chrome devtools, most of the extra memory is allocated to objects and arrays, which are data stored in weak maps (Mavo.elementData, Mavo.Node.elements, Mavo.DOMExpression.elements, Bliss.listeners). Most confusingly, when checking the weakmap in console, the reported references are not longer there. For example, devtools may report Mavo.elementData 56 as the retainer of some objects, even though Mavo.elementData has less than 56 entries at the end of the recording. I am guessing that the values of weakmaps cross reference each other, or are referenced somewhere else, and thus prevent them from being garbage collected, even after deleting the keys from the weakmaps.

kind regards.

Oliver

PR #422, opened 2018-12-10
make the defined property of Mavo.all configurable

hopeful2:
As requested in issue #410, there are use cases to activate Mavo manually and destroy instances when done. This makes it much faster to load/render a page with multiple Mavo apps not activated yet, and activate/deactivate apps dynamically on demand. With this approach we can avoid the spike of memory consumption when loading the page, and run the apps even on older smart phones.

Currently the defined property of Mavo.all holds immutable reference to created instances,  and prevents Mavo instances from garbage collection. Making the defined property configurable empowers Mavo to play a key role in larger apps. Please consider this change,

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -263,7 +263,7 @@ var _ = Mavo.Functions = {
 		}
 
 		if (num < 10 || num > 20) {
-			var ord = ["th", "st", "nd", "th"][num % 10];
+			var ord = ["th", "st", "nd", "rd", "th"][num % 10];
 		}
 
 		return ord || "th";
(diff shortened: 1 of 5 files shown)


LeaVerou:
I'm fine with this change, but it looks like the diff in this PR has your previously merged changes too!

hopeful2:
I think it is because I merged the two accepted PRs into the master branch of my forked repo, and used that as the base for editing and creating this PR. It seems the diff is between this commit, and the commit in Mavoweb/Mavo at the point of forking. 

I will fetch from the upstream repo, checkout a fresh branch and submit a new PR for this.

PR #425, opened 2018-12-11
fresh - WIP: use Map instead of WeakMap for Mavo.elementData fixes #417

### LINKED ISSUE #417: Proposal - use Map instead of WeakMap for caching data in Mavo.elementData
In our quest to manage memory after destroying a Mavo instance, we have found that WeakMap is not effective in dealing with circular references. We document our finding below for your consideration of switching to Map to manage elements caching explicitly.

Currently Mavo.elementData is a WeakMap. When Mavo runs, it uses elements as key, and stores data. The data often contains references to Mavo instance/nodes/collection items, which contain reference to each other and elements. These circular references make it impossible for the WeakMap to fulfill its mission of automatically removing cached data. A single reference in data will prevent the whole bunch from garbage collection. The elements become detached but still occupy memory after they are removed from DOM. This problem is not apparent when an app starts, however it may cause issues when the app has run for a while. 

A data structure using Map does not solve circular reference per se. However developers can iterate through the members of a Map and remove unwanted items periodically.  This is not possible with WeakMap, as by design WeakMap is not iterable. 

Mavo.Node.elements and Mavo.DOMExpression.elements are both WeakMap. Since Mavo tracks instances of Mavo.Node and Mavo.DOMExpression separately, developers can loop through the instances and remove unwanted elements from these two WeakMaps. 

Your thoughts?

hopeful2:
This is a PR based on fresh checkout of the master branch from mavoweb/mavo.

As documented in issue #417 , WeakMap is ineffective in managing circular references. It prevents manual management as by design it does not allow iteration. A data structure of Map is more suitable in this case. Please consider.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -195,7 +195,7 @@ var _ = $.extend(Mavo, {
 		return ret;
 	},
 
-	elementData: new WeakMap(),
+	elementData: new Map(),
 
 	/**
 	 * Get node from path or get path of a node to an ancestor


hopeful2:
Upon second thought, I think we can still use WeakMap for Mavo.elementData, by changing the way to set and get values. As long as we avoid circular reference, WeakMap is very helpful. I will write about the new approach in issue #417, and will close this PR.

PR #427, opened 2018-12-12
skip setting loading attribute at init stage if instance exists

hopeful2:
Upon page load, when some apps are activated manually but asynchronously using a Promise, the manual activation could happen before or after the main activation script(also a Promise) runs. If the manual activation happens first, the main activation script resets the status to loading, finds nothing to activate, exits, but leaves the status as loading. The apps function normally otherwise but the loading indicator is always shown. End users may think this is a bug, as if the loading never finishes.

This issue is very difficult to solve using a plugin or hook as the main app flow won't call it if the apps have already been activated. Hence this PR for delaying the ui update in init script. please consider. 

The status will be set later by individual functions.

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -853,7 +853,7 @@ requestAnimationFrame(() => {
 	);
 
 	_.inited = $.ready().then(() => {
-		$.attributes($$(_.selectors.init), {"mv-progress": "Loading"});
+		//$.attributes($$(_.selectors.init), {"mv-progress": "Loading"});
 		return _.ready;
 	})
 	.catch(console.error)
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -853,7 +853,12 @@ requestAnimationFrame(() => {
 	);
 
 	_.inited = $.ready().then(() => {
-		//$.attributes($$(_.selectors.init), {"mv-progress": "Loading"});
+		$$(_.selectors.init).forEach(function(elem) {
+			// skip if an instance has been created, for example by another script.
+			if (!Mavo.get(elem)) {
+				elem.setAttribute("mv-progress", "Loading");
+			}
+		});
 		return _.ready;
 	})
 	.catch(console.error)


LeaVerou:
The reason this is there is so that there's a loading indicator while plugins and dependencies are loading, which can sometimes be long, depending on the number of dependencies that needs to be loaded and network speed.

I'm not sure I understand what the issue is with the manual activation script (even after reading your message like 5 times), perhaps you can elaborate a bit on that? We can probably find a different solution, but I need to understand what the problem is first!

hopeful2:
Apology if the message was not clear. 

In our use case this is what happens when page loads, 
fetch data from local indexDB with a Promise,
render to string by section (from an array, using third party template system)
replace placeholder with mv-app attribute for the first 10 sections
concatenate to a big string
insert to DOM
look for sections in the DOM with mv-app attribute
create Mavo instance by calling new Mavo(section), if it has not been created yet.

By "manual app activation" I meant creating Mavo instances manually.  

The above runs in parallel to Mavo.init when page loads. The timing of the two promises can not be predicted. If Mavo instances have been created manually, Mavo script resets the status to loading, finds nothing to do, but exits leaving the status as loading. The loading indicator always stay in the UI.

We could introduce a deliberate delay in our Promise process, and avoid competing with the main Mavo process. The problem is we are not sure for how long the delay should be.

hopeful2:
Thank you. We have already done that check, before creating instances manually.

A new idea suddenly came to my mind while reading your comments. I remember Mavo has a promise when data loading has finished. In our promise we can just fetch data, cache objects in memory, render to DOM, and return that Mavo promise. We will create instances manually if Mavo has already finished initializing and resolved that promise. otherwise we will just wait for Mavo to create the instances. This will solve the problem of unpredictable timing.  

So this UI indicator can stay as it is now. I think returning a Mavo promise should work. I will run a few tests to confirm that.

hopeful2:
Regarding indexedDB adaptor, I will prepare a plugin if you think that is useful to others. At the moment we use PouchDB for the possibility of replicating with a CouchDB server. Another plugin called CouchDB is already using PouchDB.  Our use case is different and users can use it purely as an offline web-based editor, without sending the data anywhere.

hopeful2:
I see what you meant now. I read your message sent through email, and missed that piece of edit.

I think it makes sense to have the code in the library as you suggested. I have modified and commited the code and changed the title of this PR to reflect what this is about. I have tested and confirmed that the change has solved the UI indicator problem.

Thank you again for the prompt response and guidance. Highly appreciated.

LeaVerou:
Thanks, merged!

PR #429, opened 2018-12-13
empty node id from array upon delete fixes #426

### LINKED ISSUE #426: refactoring of node.uid  and destroy method
## current status
Currently all node instances are recorded in an array Mavo.Node.all. A node instance uses its position in the array as uid, and when the node instance is destroyed, the array fills this position with null. 

## pros
quick to access child nodes by array index. this is very helpful if we record this position and relative path to other objects, and use this information for the getter function of value object stored in Mavo.elementData. This will avoid accessing a nested object using property path, from root node all the way to the great great great ... grandchild node. 

## potential issues
When nodes are destroyed by batch, and new ones added by batch, the array Mavo.Node.all will grow in size and may end up in a very long array with many null items. this will potentially slow down the operations when we need to loop through the array in the future.  This is not an issue if we loop through all nodes of an app by the node.walk method. When there are many apps on the page, walking within an app probably takes less time than looping through the array of all nodes on the page.

The null items occupy extra memory.

## alternative approach 1
use a dictionary instead, with random string as node id. 

## alternative approach 2
* take the destroyed node out of the array by splicing, without null items
* use a random string as node uid to avoid id clashes. This can be generated with Math.random with a prefix "n" for node. 

## impact
A search in the code base indicates that uid is only used in src/node.js, for accessing the node by uid to replace it with null, and toString function. 

If you think the alternative approach suits the majority of users better, I will submit a PR for this. Let's weigh the pros and cons before jumping into action.

hopeful2:
as discussed, empty the position in array held by a node upon deleting the node, but still keep the array length.

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -132,7 +132,7 @@ var _ = Mavo.Node = $.Class({
 			this.itembar.destroy();
 		}
 
-		_.all[this.uid] = null;
+		delete _.all[this.uid];
 	},
 
 	getData: function(o = {}) {


LeaVerou:
Thank you, merged!

PR #430, opened 2018-12-15
add Mavo.destroy method and use Map for Bliss.listeners

hopeful2:
This PR depends on being able to iterate through a Map of Bliss.listeners, and unbind/remove elements with mv-ownerapp attribute. This will allow WeakMaps to release resources for GC.

A revised version of Bliss Shy has been included in the lib.

I have tested and verified on Chrome that the PR has resolved memory leak issue when destroying instance and elements. More tests are welcome from other developers using different operating system and browser versions.


hopeful2:
added hooks to Mavo.destroy() method, and added Collection.destroy() method to remove reference to dragula. Thus it is no longer necessary to walk through nodes again for this purpose in Mavo.destroy().

hopeful2:
Thanks for the prompt feedback. no rush. Take your time to review using Map for Bliss.listeners. A registered listener always contains reference to the element and callback function, thus WeakMap can do nothing until the listener is removed manually. 

This PR shows what elements need attributes to identify apps, to answer your question in issue #417. It is easier to show the code using diff tools, rather than describe about the code. It shows an overview of expected changes. 

I can break it down into small changes and several PRs. A heavy storm yesterday has heavily damaged the local infrastructure of the area I live in causing blackout. I have lost internet connection and am writing this message with mobile connection. It is hard to work with mobile. I will prepare new PRs in the new year, if not before Christmas holiday.

wish you a happy and peaceful festival season.

LeaVerou:
Hi there @hopeful2,

It's been a while! Hope you're well. :)

It looks like this PR has gotten very out of date.
I seem to recall that it required a fairly substantial change to Bliss which I was not comfortable with, but there's a lot of good stuff that never got merged. 
Are you still interested in breaking it down into smaller PRs, given that it would require substantial work rebasing it to be up to date with the recent Mavo code?
If so, feel free to reopen it (or open a new one).

In the meantime, I'm going to push a simple destroy() method, since I've realized recently that Mavo instances don't have one at all.

hopeful2:
Hi @LeaVerou, thanks for looking into this issue. Sorry for the late reply. I hope you are doing well.

If I remember correctly, the problem was caused by a reference held by dragula. Once the dragula instance was destroyed, the memory issue was gone and the WeakMap was able to clean up as intended. It seems better to keep using WeakMap for its merits, as long as a reference is not held somewhere. Thus the majority of the pull request is no longer necessary. 

Please keep up the great work. I have always been amazed by the low memory usage by the project.

PR #432, opened 2018-12-19
add destroy method to Collection class

hopeful2:
with reference to discussions on issue #417, added a destroy method to Collection class and manually remove dragula instance to break circular reference.


LeaVerou:
Merged, thanks!

PR #435, opened 2019-01-09
prevent infinite recursion when registering domexpression (fix #434)

### LINKED ISSUE #434: infinite recursion bug when registering cross-app domexpression with multiple external apps
While it is fine to access values from another Mavo app in Mavo expression, trying to access values from multiple external Mavo apps will cause infinite recursion and abort after reaching maximum call stack. The reason is that `mavo.expressions.register()` function keeps calling each other app's `expressions.register()` function and then calls itself recursively when there are multiple external apps involved.

The solution is to track and check if a Mavo app has already registered the `domexpression` instance, and make sure to call `expressions.register()` only once per app. I will submit a PR for this shortly for you to consider.

hopeful2:
This PR tracks Mavo app ids and checks whether a Mavo app has already registered a domexpression instance before registering it again. This makes sure each app will register a domexpression once only, and prevents infinite recursion when an expression involves multiple external apps.

### DIFF
--- a/src/expressions.js
+++ b/src/expressions.js
@@ -24,7 +24,7 @@ var _ = Mavo.Expressions = $.Class({
 
 	register: function(domexpression) {
 		var ids = this.identifiers;
-
+		domexpression.registeredApp = domexpression.registeredApp || new Set();
 		domexpression.identifiers.forEach(id => {
 			if (!(ids[id] instanceof Set)) {
 				ids[id] = new Set();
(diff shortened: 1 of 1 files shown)


LeaVerou:
Merged, thanks!!

As a minor comment, the comment `// Cross-mavo expressions, make sure to track app id before calling register.` didn't make sense to me at first, hence why I didn't understand what the bug was about.

I'd rephrase to something like "Only register expression with app if it has not already been registered. This can happen if an expression references 2 or more different apps (see #ISSUE-ID)"

hopeful2:
Great! another bug removed. 
Yes your comment is easier to read. My original comment was trying to remind myself that the revised function should check the app id before calling the register() function in other apps.  The order of execution is important here.

PR #437, opened 2019-01-09
make sure domexpression exists before unregister recursively fixes #436

### LINKED ISSUE #436: "Uncaught ReferenceError: domexpresssion is not defined" when unregistering domexpresssion involving multiple external apps
Now that I have got `mavo.expressions.register()` working in local dev version after applying patches in PR #435 , the following error was thrown when unregistering `domexpression` involving multiple external apps. 

```
Uncaught ReferenceError: domexpresssion is not defined
at domexpression.identifiers.forEach.id
at Array.forEach (<anonymous>)
at e.unregister
... 
```

It seems when there are multiple external apps, the `domexpresssion` object is destroyed during the loop, and by the time a certain external app's `expressions.unregister(domexpresssion)`  is executed, this object no longer exists. 

I will submit a PR shortly to make sure this object still exists before calling external app's `expressions.unregister` function.

hopeful2:
When an expression involves multiple external apps, domexpression could have been destroyed during the loop when another app is destroyed. This PR makes sure domexpression still exists before calling  an external app's `expressions.unregister()` function, and thus avoids Uncaught ReferenceError thrown by browsers.

### DIFF
--- a/src/expressions.js
+++ b/src/expressions.js
@@ -48,7 +48,7 @@ var _ = Mavo.Expressions = $.Class({
 			}
 			// just in case domexpresssion has been destroyed by another app during the loop
 			// when another app is destroyed.
-			if (id in Mavo.all && 'object' == typeof domexpresssion && domexpresssion) {
+			if (id in Mavo.all && 'undefined' !== typeof domexpresssion) {
 				// Cross-mavo expressions
 				Mavo.all[id].expressions.unregister(domexpresssion);
 			}
--- a/src/expressions.js
+++ b/src/expressions.js
@@ -48,7 +48,7 @@ var _ = Mavo.Expressions = $.Class({
 			}
 			// just in case domexpresssion has been destroyed by another app during the loop
 			// when another app is destroyed.
-			if (id in Mavo.all && 'undefined' !== typeof domexpresssion) {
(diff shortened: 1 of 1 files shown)


LeaVerou:
I tried to investigate this bug and it's very bizarre indeed: `domexpression` exists just fine as execution enters the `domexpression.identifiers.forEach` loop and becomes completely undefined immediately after!

I'm ready to merge this PR, but I think the check can be shortened to `typeof domexpression !== "undefined"`. If you agree, can you do that? Thanks!

hopeful2:
That is great. I have revised the condition check as you have suggested and pushed to git.

LeaVerou:
Sorry, very minor nitpick: We only use double quotes in Mavo. There's an ESLint file in the repo to enforce code style and we generally encourage contributors to use it.
Also, I think comparisons tend to be easier to read with the constant as the second operand (i.e. `typeof domexpression !== "undefined"` instead of `"undefined" !== typeof domexpression`).

hopeful2:
No problem. `"undefined" !== typeof domexpression` is difficult to read, but it is easier for computer to catch typos such as `"targetvalue" = myVar` in JavaScript when one intends to compare the value.

LeaVerou:
I misunderstood your last comment and thought you hadn't made the change. Just noticed that you had. Merged, thank you so much!

PR #448, opened 2019-01-28
Fix MavoScript operator precedences

betaveros:
Make `==` and `!=` the same precedence as the comparison operators, so
that they chain properly.

Also make `mod` the same precedence as `*` `/` and `%`, which I think is
the only sensible precedence for it; it might have been an error
introduced in babf794e79b66dfef947af97e0732c4856ee8518.


LeaVerou:
Thanks!

PR #450, opened 2019-01-28
Only use compare for heterogeneous chained comparisons

betaveros:
Simplify generated code for chained comparisons when possible, whenever
only one comparison is used, which will cover the most common uses and
matches the old behavior.


LeaVerou:
Wow, that was fast! Thanks!

PR #467, opened 2019-02-04
Make comparisons with an undefined operand return true

### LINKED ISSUE #404: Comparing lists of mismatched lengths extends with `true` (approximately, 1)

betaveros:
After the last few PRs I think comparison operators behave sufficiently
differently from other operators that it's not worth it to specify their
behavior with the same attributes other operators use, such as like
default.

(Actually) fixes #404.

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -36,7 +36,10 @@ var _ = Mavo.Script = {
 				var rightDefault = o.rightDefault === undefined ? o.default : o.rightDefault;
 
 				for (let i = 0; i < max; i++) {
-					if (a[i] === undefined) {
+					if (o.comparison && (a[i] === undefined || b[i] === undefined)) {
+						result[i] = true;
+					}
+					else if (a[i] === undefined) {
 						result[i] = rightUnary ? rightUnary(b[i]) : o.scalar(leftDefault, b[i]);
 					}
 					else if (b[i] === undefined) {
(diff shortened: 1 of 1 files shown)


betaveros:
I'm not totally sure what you mean. Would you prefer the following (which ends up duplicating both `a[i] === undefined` and `b[i] === undefined`)? I don't see any way to write this logic that doesn't duplicate a condition.

```js
if (o.comparison && (a[i] === undefined || b[i] === undefined)) {
	result[i] = true;
}
else if (a[i] === undefined) {
	result[i] = rightUnary ? rightUnary(b[i]) : o.scalar(leftDefault, b[i]);
}
else if (b[i] === undefined) {
	result[i] = leftUnary ? leftUnary(a[i]) : o.scalar(a[i], rightDefault);
}
else {
	result[i] = o.scalar(a[i], b[i]);
}
```

LeaVerou:
It's better to duplicate a condition than to duplicate **both** a condition and the code to be executed, so I'm finding your new version better.

I can't think of a way to not duplicate anything, and neither could @karger for now, so if you can make the change your suggest in your last comment, I think it's ready to merge. :)

One way to only duplicate one of the two conditions would be this, but I wouldn't say it's a significant improvement:

```js
if (a[i] !== undefined && b[i] !== undefined) {
	result[i] = o.scalar(a[i], b[i]);
}
else if (o.comparison) {
	result[i] = true;
}
else if (a[i] === undefined) {
	result[i] = rightUnary ? rightUnary(b[i]) : o.scalar(leftDefault, b[i]);
}
else {
	result[i] = leftUnary ? leftUnary(a[i]) : o.scalar(a[i], rightDefault);
}
```

LeaVerou:
Hey, it's always a good practice to post a comment once you're done with changes so that a) I know that you're not planning to push any more commits and I should review and b) because Github doesn't always notify me (e.g. I had no idea, I just happened to check the PR by coincidence).

Merged now, thanks!!

PR #480, opened 2019-02-13
Make most comparisons return undefined on one undefined operand

### LINKED ISSUE #404: Comparing lists of mismatched lengths extends with `true` (approximately, 1)

betaveros:
Uses the "default" property to mean the result of the comparison when
one element is undefined. Also make the "empty" function say that
undefined is falsy. Fixes #404.

(The amount of changes to handle `undefined` is a little large for my tastes. Also I see that `empty` tested for `null` before this PR. Does Mavo use `null` right now? I think we definitely do not want to handle both `null` and `undefined` in Mavo.)

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -570,7 +570,7 @@ function str(str = "") {
 
 function empty(v) {
 	v = Mavo.value(v);
-	return v === null || v === false || v === "";
+	return v === undefined || v === null || v === false || v === "";
 }
 
 function not(v) {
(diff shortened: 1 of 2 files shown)


LeaVerou:
Yes, Mavo does use `null` in a number of places to mean "empty".
I agree that we probably don't want to handle both `null` and `undefined`. 
The amount of changes is not dramatic (I've seen way larger PRs), but props for being sensitive to that! 
Like I said, I'm open to just returning `false` in these cases.
Would it be too much trouble to try that in another branch and submit another PR? I would love to see what is the cost of returning `undefined`, but I'm learning towards just returning false and be done with it, and based on your comment I think you agree too (?).

PR #482, opened 2019-02-14
Make most comparisons return false on one undefined operand

### LINKED ISSUE #404: Comparing lists of mismatched lengths extends with `true` (approximately, 1)

betaveros:
Uses the "default" property to mean the result of the comparison when
one element is undefined. Fixes #404. (Alternative to #480.)


LeaVerou:
Yup, I think that's much better. Let's go ahead with this.
Thank you so much for the fast turnaround!!

PR #510, opened 2019-03-16
Bug: contains function

efeichen:
Addressed some issues after converting `search()` function with `binaryOperation()`. All tests pass except:

1. The ones where the needle is an object. Do you think it'll be a good idea to enable the function to check if a haystack contains a specified object?
2. An edge case where either the haystack or needle is `[]`. Instead of defaulting to `false`, the function will return `[]` because when `binaryOperation` is called, the arguments won't be processed by the scalar since `[].length = 0`. This seems rather ungraceful to me, especially in the case of `contains` and `search` where it's quite likely that Mavo apps will search an empty list and the developer will expect the function to return `[false]`. Perhaps in `binaryOperations` I could check if this edge case appears, do `scalar(leftDefault, rightDefault)`?

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -445,38 +445,34 @@ var _ = Mavo.Functions = {
 	}),
 
 	/**
-     * Search if a group, collection, or primitive contains needle
-	 * @returns Boolean if a haystack of object or primitive is passed
-	 * @returns Array of booleans if a haystack of array is passed
-     */
-    contains: (haystack, needle) => {
-		var ret = Mavo.Script.binaryOperation(haystack, needle, {
-			scalar: (haystack, needle) => {
-				if ($.type(haystack) === "object") {
-					for (var property in haystack) {
-						ret = _.contains(haystack[property], needle);
-						if (Array.isArray(ret)) {
-							ret = Mavo.Functions.or(ret);
-						}
-						if (ret) {
-							return true;
-						}
-					}
-				}
-				else {
-					return _.search(haystack, needle) >= 0;
-				}
-				return ret;
-			},
-		});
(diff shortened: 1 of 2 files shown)


LeaVerou:
Merged, thanks!
Btw using `ret` for the name of the variable that contains the returned result is a pretty widespread convention, both in our codebase and elsewhere. But I didn't want to hold up the PR for that.

About your questions:

1. Given that two equal objects will be serialized the same way, I wonder if we could easily implement this by just string searching the JSON serialization of both needle and haystack. However, that might end up being too strict. For example, we probably want to ignore the order of properties, whereas this will force a specific order. Also, we might want to search for objects that are equal or a superset, and that won't allow for this either. But it may be better than nothing, and these things may even be possible with some clever use of the `replacer` argument of `JSON.stringify()`. 
2. I'm not sure `[]` is that wrong when the haystack is `[]`, since something like `contains(['foo'], ['bar'])` would return `[false]`, and they are different cases. However, the needle being `[]` is not the same as the haystack being `[]`. I'm not sure how we should best handle that, I'll think about it.

efeichen:
Hey Lea, I want to bring this PR back up since we resolved the first issue about object type needles but never really came to a consensus for the second issue. Should we delve more into handling this edge case gracefully?

LeaVerou:
Hi Kenneth,

Could you please remind me what the issue was or point me to a comment that discusses it? It's been a while and I've forgotten!
Sorry about that!

efeichen:
Ya of course! There's a failing `contains()` test where `contains([], "foo")` would return `[]` while I set that the expected return is `false`. We are not sure whether this is the best result, so some discussion is needed.

The bulk of the discussion is at the start of this PR.

PR #522, opened 2019-04-15
Add .mv-no-hiding-during-loading class

betaveros:
Add a class that can be used to suppress the hiding of mv-multiple
elements when loading. Useful for the upcoming server-side rendering.

### DIFF
--- a/src-css/mavo.scss
+++ b/src-css/mavo.scss
@@ -234,7 +234,7 @@ button.mv-close {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
LGTM, thx!

karger:
would .mv-show-during-load be less of a mouthful?

LeaVerou:
> would .mv-show-during-load be less of a mouthful?

Hmm, maybe `mv-visible-load`?

PR #552, opened 2019-10-13
Fix redirects issue

DmitrySharabin:
I've played with a couple of similar rules on my repo. That should work.
There is another one issue I am going to fix: in the repo, we don't have releases prior to 0.1.3, so we should fix the generator.


LeaVerou:
Wow, nice, thanks!

PR #556, opened 2019-10-13
Fix newly found redirect issues

DmitrySharabin:
As redirect rules resolve from top to bottom, we need to define more specific rules earlier in the list. We also need a separate rule for the addresses like v0.2.x., because they are (surprisingly) are not covered by the existing /v0.2/* rule—the address v0.2.0 is not redirected.


LeaVerou:
Thank you!

PR #562, opened 2019-12-19
Make <select> read-only, fixes  #559

### LINKED ISSUE #559: Firefox shows editing popup for dynamic <select> if <option> has the `mv-attribute="value"` attribute
The reduced testcase: https://codepen.io/dmitrysharabin/pen/LYENWEb

OS: macOS Catalina (10.15.1)
Browser: Firefox 71.0 (64-bit)

I also noticed that in Chrome, when I change the value, I can see the intermediate values between the previous value and the new one (e.g., if the value was 2 and I change it to 9, I notice that the final value sets not immediately).

DmitrySharabin:
> ... it doesn't quite make sense to show an editing popup for <option> but primarily because it should be read-only!

If I am not mistaken, not only `<option>` should be read-only, but `<select>` itself.

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -217,6 +217,7 @@ _.register({
 	"select": {
 		extend: "formControl",
 		selector: "select",
+		modes: "read",
 		subtree: true
 	},
 


LeaVerou:
But `<select>` is edit-only, like all form elements on their default attribute. And it can be read-write on other attributes. For `<option>` I agree, there's no sensible way to make it read-write, nor is it editable.

DmitrySharabin:
Oh! That is how the `modes` property work. I’m sorry, I didn’t get that earlier. 😔
Will it be enough simply add something like that in `elements.js`?
```"option": {
		modes: "read"
	},
```
Or I need to look more carefully through the codebase to find more appropriate place?

DmitrySharabin:
Close this PR because I mislocated the problem. I will try to solve the problem in the right way.

PR #566, opened 2020-01-04
Revert "Make <option> read-only, fixes 559"

LeaVerou:
Reverts mavoweb/mavo#564

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -220,12 +220,6 @@ _.register({
 		subtree: true
 	},
 
-	"option": {
-		extend: "formControl",
-		selector: "option",
-		modes: "read"
-	},
-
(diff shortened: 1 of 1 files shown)


LeaVerou:
@DmitrySharabin you could try taking another stab at fixing the original issue if you want :) 
I'd suggest using Requestly to alias all Mavo URLs to your local URLs and browsing the Mavo site, running the testsuite and any other Mavo apps you can think of, to make sure no regressions are introduced.

DmitrySharabin:
I will definitely take another stab at fixing this issue—I need to “beat” it. :)
And thanks for telling me about Requestly—I’ve never heard of this useful extension.

Issue #573, opened 2020-02-25
<select multiple> doesn't work


LeaVerou:
Reduced testcase: https://codepen.io/leaverou/pen/gOpgjgr?editors=1100

Should be easy to fix.

DmitrySharabin:
Tried to fix this. Please have a look.

PR #578, opened 2020-04-04
Make <select multiple> work

DmitrySharabin:
Fix for #573

Registered the new `select[multiple]` element.

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -224,10 +224,12 @@ _.register({
 		extend: "select",
 		selector: "select[multiple]",
 		getValue: element => {
-			return Array.from(element.selectedOptions).map(option => option.value);
+			return Array.from(element.selectedOptions).map(option => option.value).join();
 		},
 		setValue: (element, value) => {
-			value = Array.isArray(value)? value : [value];
+			// Why +""? If the value is being set via mv-value and is a number,
+			// we must convert it to a string to avoid extra checks.
+			value = Array.isArray(value)? value : (value + "").split(/\s*,/);
 
 			Array.from(element.options).forEach(option => {
 				// Why? If the value is being set via mv-value,
--- a/src/elements.js
+++ b/src/elements.js
@@ -229,7 +229,7 @@ _.register({
 		setValue: (element, value) => {
 			value = Array.isArray(value)? value : [value];
 
-			$$("option", element).forEach(option => {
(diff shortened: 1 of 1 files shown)


src/elements.js
```diff
@@ -220,6 +220,39 @@ _.register({
...
 	},
 
+	"select[multiple]": {
+		extend: "select",
+		selector: "select[multiple]",
+		getValue: element => {
+			const selected = element.selectedOptions;
+			const value = [];
+
+			if (selected.length) {
```
LeaVerou, line 230: I think just `Array.from(element.selectedOptions).map(option => option.value)` should work, but I haven't tested it.

src/elements.js
```diff
@@ -220,6 +220,39 @@ _.register({
...
+					value.push(option.value);
+				}
+			}
+
+			return value;
+		},
+		setValue: (element, value) => {
+			value = Array.isArray(value)? value : [value];
+
+			$$("option", element).forEach(option => {
```
LeaVerou, line 241: Slightly preferable to use `element.options` instead of a new DOM query.

src/elements.js
```diff
@@ -220,6 +220,39 @@ _.register({
...
+			if (selected.length) {
+				for (const option of selected) {
+					value.push(option.value);
+				}
+			}
+
+			return value;
+		},
+		setValue: (element, value) => {
+			value = Array.isArray(value)? value : [value];
```
LeaVerou, line 239: I'm not sure about expecting an array as a value on a primitive. In most cases, it will not be an array. I would suggest expecting a comma-separated string (convert to array via `.split(/\s*,/)`), but also handling arrays, for the best of both worlds.

src/elements.js
DmitrySharabin: Yes, it works. Thank you. This is much much clearer.

src/elements.js
DmitrySharabin: You are absolutely right!
I must examine the DOM API more thoughtfully. 😞

src/elements.js
DmitrySharabin: Fixed that. I even caught a bug, thanks to your comment. Thank you.

DmitrySharabin:
Fixed the issues you pointed out. Thank you.

LeaVerou:
Merged, thanks!

PR #579, opened 2020-04-04
Allow specifying different upload backend than storage backend

### LINKED ISSUE #537: Allow specifying different upload backend than storage backend
E.g. someone using local storage might still want to upload images to Github. Or there could even be plugins for backends that are only suitable for uploads, e.g. Imgur.

Implementation details:

The code already makes a distinction between the storage backend and the upload backend (however `this.uploadBackend` is currently just a getter that returns the storage backend IFF it has an upload method), so this should be rather easy to implement.

DmitrySharabin:
Let's try to close #537 and make Mavo even more powerful.

I suggested that this type of backend can be specified via the `mv-upload-backend` attribute (because `mv-uploads` and `mv-upload-url` are already taken, and `mv-upload` can confuse users).

I could miss some key points or messed something up, so I open to critique.

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -165,7 +165,7 @@ _.register({
 		editor: function() {
 			var kind = this.element.nodeName.toLowerCase();
 			kind = kind == "img"? "image" : kind;
-			Mavo.setAttributeShy(this.element, "mv-uploads", kind + "s");
+			Mavo.setAttributeShy(this.element, "mv-upload-path", kind + "s");
 
 			return this.createUploadPopup(kind + "/*", kind, "png");
 		}
@@ -176,7 +176,7 @@ _.register({
 		attribute: "href"
 	},
 
-	"a[mv-uploads], link[mv-uploads]": {
+	"a[mv-upload-path], link[mv-upload-path]": {
 		default: true,
 		attribute: "href",
 		editor: function() {
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -85,7 +85,7 @@ var _ = self.Mavo = $.Class({
 
 		this.permissions = new Mavo.Permissions();
 
-		var backendTypes = ["source", "storage", "init"]; // order is significant!
+		var backendTypes = ["source", "storage", "init", "upload-backend"]; // order is significant!
 
 		// Figure out backends for storage, data reads, and initialization respectively
 		backendTypes.forEach(role => this.updateBackend(role));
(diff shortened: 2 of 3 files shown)


LeaVerou:
Hi Dmitry,
Thanks for working on this! Looks ok, though I would *really* prefer a single word name. Thinking what it could be. 

Have you tested it? Does it work?

DmitrySharabin:
Yes, I have tested it, and it worked. And it didn't break any tests we already have.

DmitrySharabin:
I have. You can't even imagine how happy I am that it works! 🥳

LeaVerou:
I would suggest using the name `mv-uploads` and changing the current `mv-uploads` to `mv-upload-path`, what do you think? It's not a very used attribute, so I think backwards compat is not a biggie.

DmitrySharabin:
Totally agree.

LeaVerou:
Wanna make the change? It will need a change in the docs too, and any demos (though I doubt we're using `mv-uploads` in any demos, I think I'm only using it on forkgasm lol

DmitrySharabin:
Sure. You can rely on me. :)

DmitrySharabin:
> I think I'm only using it on forkgasm lol

And on haystack/website also. :)

winniejo:
so is it mv-uploads or mv-upload-path ?

DmitrySharabin:
Hey @winniejo! To specify upload backend use the `mv-uploads` attribute on the root element (with the `mv-app` attribute). But I'm not sure whether we built the dev version of Mavo you can play with (although the feature has been merged). We need to ask Lea (@LeaVerou) about it.

For details on how to use `mv-upload-path`, [see the docs](https://mavo.io/docs/storage#uploads). We've already updated it.

LeaVerou:
The `get.mavo.io/mavo.js` URL is updated automatically with every commit :) 

@DmitrySharabin did we add `mv-uploads` to the docs? If not, could you add it please? Make sure to mention it is only available in the development version (and let's make a note to update that when we release a new version)

winniejo:
Okidoki  Lea Verou thnks

Joe

DmitrySharabin:
@karger 

I might be mistaken but to my knowledge `mv-storage!= mv-source+mv-uploads`. `mv-source` only specifies from where we load our data, `mv-uploads`—where we put our assets (not data). So without specifying `mv-storage`, we can't store our data anywhere. That's why providing `mv-source` and `mv-uploads` is not enough to cover the functionality of `mv-storage`.

If we provide both `mv-storage` and `mv-uploads` and both of them have the `upload` method, then `mv-uploads` wins. In any other case, wins the one who has the `upload` method, or the upload feature is not provided to an end-user at all.

DmitrySharabin:
@LeaVerou 

> If not, could you add it please? Make sure to mention it is only available in the development version (and let's make a note to update that when we release a new version)

Done. I wasn't sure where to put the note to make it available for all of us, so I made a [wiki page](https://github.com/mavoweb/mavo/wiki/To-Update-in-the-Docs).

PR #583, opened 2020-04-18
Improve error reporting

### LINKED ISSUE #101: Custom error messages
Post-release leftover work from #99: Ability to include a custom error message template, with class mv-error.

DmitrySharabin:
Let's try to close #101, and add to Mavo the ability to Include a custom error message template, with class `mv-error`.

I need to put into words a couple of thoughts I have to make sure I get everything correctly:

1. We can provide only one error message per app. Even though we have three potential cases when an error may occur: while loading, saving, or uploading. Should we distinguish these cases somehow?

2. We should provide a custom error template inside a `hidden` wrapper element in order not to render it on a page before we need it, like so:
```html
<div hidden>
  <div class="mv-error">
    <strong>Error!</strong> We have a problem. Check the browser console for details.
  </div>
</div>
```

Are we OK with that?

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -326,8 +326,10 @@ var _ = self.Mavo = $.Class({
 		return new _.UI.Message(this, message, options);
 	},
 
-	error: function(message, ...log) {
-		this.message(message, {
+	error: function (message, ...log) {
+		// If a custom error message is provided (in an element with class mv-error),
+		// show it to an end-user (Fix for #101)
+		this.message($(".mv-error", this.element) || message, {
 			type: "error",
 			dismiss: ["button", "timeout"]
 		});
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -329,7 +329,7 @@ var _ = self.Mavo = $.Class({
 	error: function (message, ...log) {
 		// If a custom error message is provided (in an element with class mv-error),
 		// show it to an end-user (Fix for #101)
-		this.message($(".mv-error", this.element) || message, {
+		this.message($(".mv-error:not(.mv-ui):not(.mv-message)", this.element) || message, {
 			type: "error",
 			dismiss: ["button", "timeout"]
 		});
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -329,7 +329,16 @@ var _ = self.Mavo = $.Class({
 	error: function (message, ...log) {
 		// If a custom error message is provided (in an element with class mv-error),
 		// show it to an end-user (Fix for #101)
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks for working on this!

Some thoughts:

- I see that it requires an element with a class of `mv-error`, and that it performs a DOM query on every error. Doesn't that mean that depending on the placement of this element, it's possible to pick up an actual error instead of the template?
- It may be a good idea to use an actual template element.

> We can provide only one error message per app. Even though we have three potential cases when an error may occur: while loading, saving, or uploading. Should we distinguish these cases somehow?

The idea is that the template provides the HTML scaffolding, but its message can vary.

DmitrySharabin:
> Doesn't that mean that depending on the placement of this element, it's possible to pick up an actual error instead of the template?

You are absolutely right. Thanks for pointing this out. I believe, using the `.mv-error:not(.mv-ui):not(.mv-message)` selector will fix that.

> It may be a good idea to use an actual template element.

Yep. I made some changes so that we can use either of the ways:

```html
<div hidden>
  <div class="mv-error">
    <strong>Error!</strong> We have a problem. Check the browser console for details.
  </div>
</div>
```

or

```html
<template class="mv-error">
  <strong>Error!</strong> We have a problem. Check the browser console for details.
</template>
```

We can point out in the docs that the second way is preferable.

LeaVerou:
I just looked at this more closely.
#101 is about providing custom HTML for error messages, not custom error message text. The error text is handled by localization. Admittedly, it's hard to tell what #101 is about since there's no real text in it, as it predates Mavo being released. It seems to me that you're reading the message itself from the HTML, so this change would make every single error in Mavo equal to the static error in the template? What's the use case for that?

DmitrySharabin:
> It seems to me that you're reading the message itself from the HTML, so this change would make every single error in Mavo equal to the static error in the template? What's the use case for that?

Right. That doesn't make sense. That was precisely my concern when I was asking about three potential cases when an error may occur and whether we need to handle them differently. Now I see that I chose the wrong file (and the wrong place) to make changes in.

Do you mind me getting another try? Sorry for forcing you to work more. I need some time to gain some experience and become more helpful. I find this way the best since I can learn directly from you. I appreciate that.

LeaVerou:
Please don't apologize, it's fantastic to have people willing to contribute. It always takes time to get acquainted with a codebase, even with experienced programmers. Don't feel bad at all!

DmitrySharabin:
Not very urgent. And I can do it better next time.

Issue #469, opened 2019-02-08
Can't control the order of the elements in the customized toolbar


DmitrySharabin:
I've faced another issue. If I try to customize the Mavo toolbar and use a custom toolbar element I can't control the order of the elements in the toolbar. Neither by changing their order in the attribute mv-bar nor by changing their order in the element I use for the toolbar. My reduced testcase is [here](
https://codepen.io/dsharabin/pen/pGajZd?editors=1100).

DmitrySharabin:
The issue is also reproduced if the relative syntax is used in the `mv-bar` attribute. Added the corresponding examples in the reduced test case above.

LeaVerou:
It seems that even `mv-bar="import export" has the same result, i.e. the order is not working there either. This looks like some sort of regression.
Is it only with import, export?

DmitrySharabin:
> It seems that even mv-bar="import export" produces the same result, i.e. the order is not working there either.

That's weird—I don't have this problem.

I tested different cases by re-arranging buttons in the `mv-bar` attribute. Indeed I can control the order of the buttons. But as soon as I enable the relative syntax, the order is out of my control. Not only with the `import` and `export` buttons but with any other (I tested all buttons mentioned in the docs: `status`, `edit`, `save`, `import`, `export`, `login`, `logout`).

LeaVerou:
I think we should discuss what *should* happen a bit.
Obviously, when we have `with import export` import should come before export. But where should they go? In the beginning? In the end? Somewhere in between?

DmitrySharabin:
We can generate many cases where users need (or simply want) to specify the order of _some_ buttons in the default set, but not all of them. In that case, I think we might retain the order of the specified buttons (no matter where they are in the default set) and keep the rest untucked.

So there will be no need to think _where_ to put the pointed buttons exactly, we just would need to reorder them in the default set if required.

That was the idea I tried to put into code in the PR. Though not very successful. lol

LeaVerou:
We cannot reorder them in the default set, because we may have multiple Mavo apps on the page, each with a different ordering. E.g. one may have `with import export` and another `with export import`. We should absolutely follow that order in the individual Mavo app though. Not sure what the best algorithm for that would be.

DmitrySharabin:
Yep. I racked my brain, trying to hit on a solution.

LeaVerou:
Don't beat yourself up, I think last time I tried to think about this, I couldn't come up with a good solution either.
[...]

PR #584, opened 2020-04-20
Fix for #469

DmitrySharabin:
Retain button order in the bar when the relative syntax is used.

Added the corresponding test to the test suite.

### DIFF
--- a/src/ui.bar.js
+++ b/src/ui.bar.js
@@ -209,7 +209,7 @@ var _ = Mavo.UI.Bar = $.Class({
 						return keep || (!_.controls[id].optional && !drop);
 					});
 
-					ids = ids.filter(id => all.contains(id));
+					ids = ids.filter(id => all.includes(id));
 
 					ids.forEach(id => all.splice(all.indexOf(id), 1, undefined));
 					ids.forEach(id => all.splice(all.indexOf(undefined), 1, id));
--- a/src/ui.bar.js
+++ b/src/ui.bar.js
@@ -209,12 +209,7 @@ var _ = Mavo.UI.Bar = $.Class({
 						return keep || (!_.controls[id].optional && !drop);
 					});
 
-					ids = ids.filter(id => {
-						const positive = all.lastIndexOf(id);
-						const negative = all.lastIndexOf("no-" + id);
-
-						return positive > Math.max(-1, negative);
-					});
+					ids = ids.filter(id => all.contains(id));
 
 					ids.forEach(id => all.splice(all.indexOf(id), 1, undefined));
 					ids.forEach(id => all.splice(all.indexOf(undefined), 1, id));
--- a/src/ui.bar.js
+++ b/src/ui.bar.js
@@ -200,21 +200,39 @@ var _ = Mavo.UI.Bar = $.Class({
 				ids = Mavo.Functions.unique(ids.reverse()).reverse();
 
 				if (relative) {
-					all = all.filter(id => {
+					// For every control in a default set, we decide whether to drop it or not and add an extra property to it—the pos property.
+					// Controls specified by a user have pos >= 0 (their relative position in a user's set).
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hi Dmitry,

Thank you so much for working on this!!

It may very well be because I'm really sleepy, but I can't understand how this change works. Could you explain and add a comment in the code too?

DmitrySharabin:
Let me explain it here first, and then we'll decide what comments to add to the code.

After detecting that we use the relative syntax, we get a set of buttons that will be available in the bar (lines 203 through 210). Their order is determined by the order of corresponding properties in the source code since the `all` variable has value `Object.keys(_.controls)` (see line 188).

But for some buttons, we need to change their order to retain the one specified by a developer. This order is provided in the `ids` variable, but since we haven't changed its value, there might be values that are not acceptable as a button's `id` (e.g., `foo`) or have already been taken into account in the previous step (were kept or dropped).

So in line 212, we drop from `ids` all the values that are not in the set of buttons in the bar (by retaining the order of the kept ones).

Then, in the full set of buttons in the bar, we “mark” positions of the buttons specified by a developer (by replacing them with `undefined`—it could be any distinguishable value)—line 214. And in the next step (line 215), we fill in “the gaps” (the marked positions) with the buttons in the order specified by a developer.

I am not sure that my explanation is clear, so I made a sketch with an example. Please, have a look: 
https://excalidraw.com/#json=5753373420683264,U9lDYzx9SaRIzxK0sGmuTQ

My drawing skills might drive you crazy. lol

DmitrySharabin:
I think I need to think it over a bit more—too many extra steps. Let's see whether I could find a more intelligent solution.

LeaVerou:
Wow, I love using a sketch to illustrate! I'm still a bit confused, but I sort of get it better now.
And yes, I think there could be a better solution (using undefined like that is a bit of a hack), but if you can't find one and you're confident that this one fixes the bug and doesn't introduce regressions, I could merge it.

karger:
I'm having trouble understanding the algorithm and the sketch so I was hoping to understand how it differs from the following simple one:

- start with default buttons
- delete all no-foo buttons
- move all specified buttons to the front, in order specified

I guess an alternative to (3) would be to move them to the back, but I can't think why that would be better (or worse).

I guess one other option to promote flexibility would be to add a "rest" keyword that indicates where all the unspecified buttons should be, so you could say e.g. 
buttons = "with import rest export"

DmitrySharabin:
David, will you please have a look at the refactored code?  I committed it recently. I’ll be grateful for any feedback.

karger:
Dmity I've looked and couldn't immediately make sense of the various position calculations and sorting that you used.  Does this design stem from a goal of being very efficient?   Because given the small number of buttons I would instead incline towards something less efficient but more transparent, such as
- from the default list, remove all buttons in the "with" list (ignoring no- prefix)
- attach at front (or end) all the buttons added in "with", in order specified.

Or perhaps your goal is not efficiency but to achieve a more appropriate ordering.  If so I'm not sure what is the characteristic of the ordering you want to achieve.

DmitrySharabin:
David, thank you so much for your feedback. I appreciate it.

You are right, I complicated the solution way too much. My main goal was to reuse the existing code as much as I can and try to keep the current order of controls in the default set and swap only some of them to retain the order of controls specified by a user. Unfortunately, I failed. But I find my attempts extremely useful. Maybe not for Mavo, but definitely for me. :)

I totally agree I must choose a more transparent way to fix the issue.

> Or perhaps your goal is not efficiency but to achieve a more appropriate ordering. If so I'm not sure what is the characteristic of the ordering you want to achieve.

I really like the idea of adding the `rest` keyword. I believe it would cover all the cases we try to cover now. Thank you.

karger:
minor thought, but perhaps instead of "rest" (which is confusing in an http context) "others" would be a better catch-all keyword.

Issue #585, opened 2020-04-25
Support async functions in expressions


LeaVerou:
Reporting on behalf of @tarfahalrashed 

Right now if a function returns a promise, all we get printed out is `[object Promise]` which is pretty much never desirable.
We should check if the return value is a Promise and update the expression when the promise resolves or rejects. If what I have in mind works, it should be a rather straight forward change in `src/domexpression.js`.

I think this would be an easy issue to tackle, @DmitrySharabin do you want to give it a shot? If not, I will work on it this weekend, but thought it would be good practice.

DmitrySharabin:
I would love to give it a shot. Thanks.

Aadi-27:
Is this issue still open or done ?

LeaVerou:
@Aadi-27 Not done. A [PR](https://github.com/mavoweb/mavo/pull/586) was opened but had some issues and was never merged.
I may give it a shot later this month, as we need it for a project.
Were you interested in working on it, or asking because you need the functionality?

LeaVerou:
In discussion about this today, I realized that what actually needs to be supported is not just async expressions, but also *async data*. So it looks like this change should be made in `Mavo.Primitive.setValue()`.

LeaVerou:
Testcaase (currently failing): https://codepen.io/leaverou/pen/dyOoVOK

LeaVerou:
I'm implementing this today and ran into some interesting dilemmas wrt to setting a property/expression to one promise, and then to another promise (let's call them promiseA and promiseB) before promiseA resolves. Let's use timed promises as an example to illustrate.

- If promiseA resolves after 5 seconds and promiseB after 1 second it seems reasonable that promiseB's resolved value gets applied after 1 second and promiseA's resolved value is never applied.
- If promiseA resolves after 1 second and promiseB after 5 seconds, do we display the value of promiseA for the 4 seconds in between or do we treat it basically the same way as if only promiseB was specified?

LeaVerou:
Also note that the current implementation only deals with promises as the final output. It does not deal with things like `promise + 1`, that will still be `NaN`. Not sure if the use cases are worth the complexity.

karger:
I ran into a use case for this today, which led me to this issue, which it appears was never resolved (sic)

LeaVerou:
> I ran into a use case for this today, which led me to this issue, which it appears was never resolved (sic)

What was the use case?

PR #586, opened 2020-04-25
Add support for async functions in expressions

DmitrySharabin:
Fix for issue #585

### DIFF
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -162,10 +162,28 @@ var _ = Mavo.DOMExpression = $.Class({
 				if (env.value instanceof Promise) {
 					env.value
 						.then(value => {
-							env.expr.function = () => value;
-							this.update();
+							env.value = Mavo.Primitive.format(value, {
+								attribute: this.attribute,
+								element: this.element
+							});
+
+							this.output(env.value);
+
+							Mavo.hooks.run("domexpression-update-end", env);
 						})
-						.catch(error => env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error));
+						.catch(error => {
+							env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error);
+
+							// Return empty string if a promise is rejected
+							env.value = Mavo.Primitive.format("", {
+								attribute: this.attribute,
+								element: this.element
+							});
+
+							this.output(env.value);
+
+							Mavo.hooks.run("domexpression-update-end", env);
+						});
 
 					// Nothing changed. We simply have a promise to work with later
 					changed = false;
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -159,6 +159,18 @@ var _ = Mavo.DOMExpression = $.Class({
 					return "";
 				}
 
+				if (env.value instanceof Promise) {
+					env.value
+						.then(value => {
+							this.expression = this.syntax.start + value + this.syntax.end;
+							this.parsed = this.syntax.tokenize(this.expression);
+						})
+						.catch(error => env.expr.error(`A promise in the expression ${this.expression} was rejected with the reason`, error));
+					
+					// Don’t print "[object Promise]"
+					return "";
+				}
+
 				return env.value;
 			}
 
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -168,7 +168,7 @@ var _ = Mavo.DOMExpression = $.Class({
 						.catch(error => env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error));
 
 					// Nothing changed. We simply have a promise to work with later
-					change = false;
+					changed = false;
 
 					// We don't want "[object Promise]" to become the expression's value
 					return "";
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -165,7 +165,7 @@ var _ = Mavo.DOMExpression = $.Class({
 							this.expression = this.syntax.start + value + this.syntax.end;
 							this.parsed = this.syntax.tokenize(this.expression);
 						})
-						.catch(error => env.expr.error(`A promise in the expression ${this.expression} was rejected with the reason`, error));
+						.catch(error => env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error));
 					
 					// Don’t print "[object Promise]"
 					return "";
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -162,12 +162,15 @@ var _ = Mavo.DOMExpression = $.Class({
 				if (env.value instanceof Promise) {
 					env.value
 						.then(value => {
-							this.expression = this.syntax.start + value + this.syntax.end;
-							this.parsed = this.syntax.tokenize(this.expression);
+							env.expr.function = () => value;
+							this.update();
 						})
 						.catch(error => env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error));
-					
-					// Don’t print "[object Promise]"
+
+					// Nothing changed. We simply have a promise to work with later
+					change = false;
+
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
Hi Lea,

Thank you for the clarification. I was sure I had been solving precisely the same task, but obviously, I was wrong. :) My mistake was that I analyzed expressions that looked like `[Promise.resolve(42)]`. And for sure, that's not every single case we can face.

I am not 100% sure that the solution I provide is correct.
It covers the following cases:

- `[Promise.resolve(42)]`
- `[Promise.reject(42)]`
- `[fetchFromAPI()]`
- `<meta property="prop" content="[Promise.resolve(42)]" /> ... <p>[prop + 1]</p>`
- `<meta property="prop" content="[fetchFromAPI()]" /> ... <p>[get(prop, "id")]</p>`
- etc.

The proposed solution doesn't cover cases like that:
- `[3 + Promise.resolve(42)]`
- `[get(fetchFromAPI(), "id")]`
- `[fetchFromAPI().title]`
- etc.

Should we cover them too?

DmitrySharabin:
I was just thinking, what if the execution of an asynchronous function will take too much time, how will end-users understand whether they need to wait a bit or that a promise has been rejected (we show it in the console)?

Should developers handle such situations themselves, or should we introduce a corresponding feature in Mavo?

I personally like the way [the Svelte framework solves it](https://svelte.dev/docs#await): rather declarative, and all three states of a promise are handled.

src/domexpression.js
```diff
@@ -159,6 +159,21 @@ var _ = Mavo.DOMExpression = $.Class({
...
+				if (env.value instanceof Promise) {
+					env.value
+						.then(value => {
+							env.expr.function = () => value;
+							this.update();
+						})
+						.catch(error => env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error));
+
+					// Nothing changed. We simply have a promise to work with later
+					change = false;
```
LeaVerou, line 171: Shouldn't this be `changed`?

src/domexpression.js
```diff
@@ -159,6 +159,21 @@ var _ = Mavo.DOMExpression = $.Class({
 					return "";
 				}
 
+				if (env.value instanceof Promise) {
+					env.value
+						.then(value => {
+							env.expr.function = () => value;
```
LeaVerou, line 165: Why are we updating the function? It seems to me that it's the return value that we need to handle specially, and leave the function as-is, but since you've worked on it you may have better insight here.

LeaVerou:
I agree it would be good to show something, e.g. "Loading..." or just a loader or whatever, however then it would trigger property updates, and we don't want that.
How about this? IF the expression is an entire element's contents, then we can add a class while the promise is unresolved, e.g. `mv-pending` or whatever. This could be a terrible idea btw, I'm just brainstorming!

@tarfahalrashed do you have any preferences on how you'd like this feature to work?

src/domexpression.js
DmitrySharabin: Absolutely right. Misspelled. Thanks!

src/domexpression.js
DmitrySharabin: My insights are the following. Please, correct me. I might be wrong.

1. On every expression update, its value (`env.value`) is a result returned by the `env.expr.function` function.
2. This function is compiled based on a parsed expression when the expression is being evaluated for the first time and is never recompiled in the future. So, if this function returns a promise, it will always be returning a promise, not its resolved or rejected value.
3. We could change this function by constructing a new expression based on data returned by a resolved or rejected promise, parse it and let Mavo compile the corresponding function for its evaluation. But the data could be anything. How can we build an expression based on an object, for example? I suppose we can't.
4. So, the only way I could find was just updating the function that computes the expression's value since its sole purpose is to return the resolved value.

LeaVerou: Indeed, but overwriting `expression.function`, while it may work, is a bit of a hack. It defeats the purpose of having a parsed function, since it needs to be re-evaluated every time.
The solution I'd go with is to check if we have a promise in DOMExpression#update(), and run the rest of its code asynchronously in those cases once the promise is fulfilled.

DmitrySharabin: Thank you very much for your patience! I think I finally got your idea. I refactored the code. I still have a couple of doubts. I'll share them with you in the comments below.

src/domexpression.js
```diff
@@ -159,6 +159,39 @@ var _ = Mavo.DOMExpression = $.Class({
...
+					env.value
+						.then(value => {
+							env.value = Mavo.Primitive.format(value, {
+								attribute: this.attribute,
+								element: this.element
+							});
+
+							this.output(env.value);
+
+							Mavo.hooks.run("domexpression-update-end", env);
```
DmitrySharabin, line 172: Should we run the hook here? I think yes, that's why I added it. But what about the `domexpression-update-start ` hook? Should we run it before changing `env.value` when the promise is resolved or rejected?

src/domexpression.js
```diff
@@ -159,6 +159,39 @@ var _ = Mavo.DOMExpression = $.Class({
...
+						})
+						.catch(error => {
+							env.expr.error(`A promise in the expression [${env.expr.expression}] was rejected with the reason`, error);
+
+							// Return empty string if a promise is rejected
+							env.value = Mavo.Primitive.format("", {
+								attribute: this.attribute,
+								element: this.element
+							});
```
DmitrySharabin, line 181: Should we print the reason why the promise has been rejected? Or is it enough to write it to the console?

Can we simply write `env.value = ""` in that case?

karger:
I'm coming to this late.  Some questions/thoughts.

What happens if a promise is rejected?

 I'm not sure we should/need to treat promises as special.  Note that mavo is already built around the idea that an expression's arguments can change at any time, causing an update to the value of an expression.  Typically that's because a variable in the expression is bound to an expression elsewhere, but having the expression be an unfulfilled promise is really no different.  Conversely, if you *do* put in special handling for not-yet-evaluated promises, perhaps we could also have the same handling for "expressions that are reactive updates based on values in the page that are taking a really long time to compute"?  I've had this experience.    

Now you might say that there's no point to doing this because by the time mavo knows an expression is in flux it knows the new value.  But if computing certain functions is slow maybe mavo can do an initial pass marking what becomes unresolved and then do a slower followup to compute new values.  Especially if they have guidance from the author:

Suppose someone writing a mavo *knows* that a particular expr will take a long time to evaluate---not because it's inherently async but because it computes a lot.  It might be nice if they could flag that---ie, wrap it in a promise.  Except our users might not know what promises are.    On the other hand, I think it's pretty easy to understand the idea that "this function might take a long time".  So how about we define e.g. a new mavo keyword called "slow" or "eventually".  slow(X) informs mavo that if X needs to be recalculated, mavo should *first* set the value of the expr to "indeterminate" and *propagate that value* (which means that any expr depending on it becomes indeterminate and that propagates too and so on).  *after* that propagation is done, mavo begins evaluation of that indeterminate expression and, once it has a value, propagates that.  This lets an author flag computations that may take long enough to demand special handling.

conversely, let's define functions resolved(X) and unresolved(X) that return booleans based on whether their argument is indeterminate.  This takes provides an alternative to lea's mv-pending because now someone can write e.g. "style=display: [if(resolved(X),visible,hidden)].

I think the idea of "no value yet" isn't too complicated, but note that we could eliminate it entirely by declaring that in mavo every expression always has a (resolved) value.  So e.g. we could define the value of mavo's fetch() to be null until the fetch completes and then update to the fetched value.  This would fit in fine because in mavo we are already used to the idea of exprs changing after we've evaluated them---so should we really force users to consider another case?

We could also give authors a construct for controlling the default value for unresolved exprs,  e.g. until(X,Y) (or perhaps "Y until X") is an expression that has value Y until X has resolved, then has value X.

DmitrySharabin:
Since now we already have this functionality in the core, I'm going to close this PR.

Issue #367, opened 2018-07-03
Moving pull request code to a plugin


LeaVerou:
I know #364 is still open about redoing the Github workflow, but since other things are prioritized above it, it could be a while before we get to it.
Meanwhile, the Github backend code has exploded in size, and I was thinking it may be a good idea to move the pull request flow to a plugin. Without the plugin, people would still be able to save their edits in forks, the plugin would just enable them to send a pull request.
That would also provide a way to disable the PR flow, which does not exist today.

@jwass91 can you look into how much code that would help us get rid of? It may turn out not to be worth doing after all.

LeaVerou:
I was reminded of this again now that I'm working on a Github Gist backend for #599 so I looked at the Github Backend code again.

It's even more imperative to do it since Jared has moved on to other things and is not maintaining this code any more and there are a bunch of bugs with it. And of course there are the issues with uploads discussed in #538. A plugin would allow users to opt-in to allowing pull-requests, since they may not always be desirable.

@DmitrySharabin any chance this is something you might want to work on? I think it's a reasonable scope and not too hard. It will require new hooks to be added, but that's easy (we can discuss together what hooks would be optimal for this, so that they are more generally useful).

DmitrySharabin:
@LeaVerou 
Sure, Lea. I would love to help. Please give me some time to have a look at the code so I could be more useful and could ask the right questions.

DmitrySharabin:
Can we close this issue now?

LeaVerou:
Sure!

PR #607, opened 2020-06-16
Remove pull request code

DmitrySharabin:
Partial fix for #367

I also added two hooks (their names are just a hint for me where exactly I put them), so we need to find suitable names for them. Which is hard, you know. 😜
All existing tests pass. But as always, I'm afraid I could miss something. 😅

### DIFF
--- a/src/backend.github.js
+++ b/src/backend.github.js
@@ -181,7 +181,6 @@ var _ = Mavo.Backend.register($.Class({
 							}
 
 							this.repoInfo = repoInfo;
-							let params = (new URL(location)).searchParams;
 
 							if (!this.mavo.source) { // if url doesn't have source, check for forks
 								if (repoInfo.fork) { // if current repo is a fork, we can display PR dialog
--- a/src/backend.github.js
+++ b/src/backend.github.js
@@ -327,7 +327,7 @@ var _ = Mavo.Backend.register($.Class({
 
 		test: function(url) {
 			url = new URL(url, Mavo.base);
-			return /^(github.com|raw.githubusercontent.com)/.test(url.host);
+			return /^((api\.)?github\.com|raw\.githubusercontent\.com)/.test(url.host);
 		},
 
 		/**
--- a/src/backend.github.js
+++ b/src/backend.github.js
@@ -38,7 +38,7 @@ var _ = Mavo.Backend.register($.Class({
 				// GraphQL
 				return this.request(info.apiCall, info.apiData, "POST")
 					.then(response => {
-						if (response.errors && response.errors.length) {
+						if (response.errors?.length) {
 							return Promise.reject(response.errors.map(x => x.message).join("\n"));
 						}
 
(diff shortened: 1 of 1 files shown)


src/backend.github.js
```diff
@@ -323,15 +184,12 @@ var _ = Mavo.Backend.register($.Class({
...
-										head: `${this.user.username}:${this.branch}`,
-										base: repoInfo.parent.default_branch
-									}).then(prs => {
-										this.pullRequest(prs[0]);
-									});
-									return repoInfo;
+								if (repoInfo.fork) { // if current repo is a fork, we can display PR dialog
+									const env = { context: this, repoInfo };
+								
+									Mavo.hooks.run("gh-login-no-source-pr", env);
```
DmitrySharabin, line 190: And here comes the second one

src/backend.github.js
```diff
@@ -147,151 +147,12 @@ var _ = Mavo.Backend.register($.Class({
...
-					username = this.repoInfo.parent.owner.login;
-					repo = this.repoInfo.parent.name;
-					base = this.repoInfo.parent.default_branch;
-				}
-				else { // Storage points to another user's repo
-					username = this.username;
-					repo = this.repo;
-					base = this.branch;
-				}
+				Mavo.hooks.run("gh-put-pr", env);
```
DmitrySharabin, line 152: Here is the first hook

src/backend.github.js
DmitrySharabin, line 190: And here comes the second one

LeaVerou:
Hi Dmitry,

Thanl you for working on this! This is a good start.

A few first observations:
- The hooks need to be related to the code they're in, not the plugin. We are not adding hooks *specifically* to make the PR plugin possible, so they cannot be named after pull requests. 
- Similarly, the remaining code cannot know about pull requests, otherwise there is no separation. 
- The UX also needs to still make sense without the plugin. People still need to be able to save forks, and be notified that they're in a fork. From a quick review of the code I can't tell for sure what happens when the plugin is not loaded. Have you tested the interaction?
- It's hard to review this without reviewing the plugin code as well. If we are going to be iterating in this PR, perhaps you could add the plugin code at the end of the file, so that they're both in one place, then once it's good to go, you move it to an actual plugin. What do you think? Alternatively, we could have two open pull requests, that works too.

DmitrySharabin:
Hi Lea,

Thanks for your comments. I went through them and worked with the code a bit more. I renamed the hooks and changed the place for one of them. So now there is not a single mention of PRs. I tested both: the backend and the plugin. The backend is fully functional without the plugin, so people still able to save forks and be notified that they're in a fork.

I sent a PR with the plugin's draft to `mavoweb/plugins`. Please have a look.

DmitrySharabin:
Here is another thing I was thinking about. Should we add mirror hooks? I mean, if we have a `gh-after-commit` hook, should we also add `gh-before-commit` (or should we call them `gh-commit-end` and `gh-commit-start` respectively)? I know it depends on whether we want devs to change some core behavior at some point and interfere into, for example, the data flow, but still.

LeaVerou:
Looks good to me, merged.
I don't think we should add hooks that we don't have use cases for.

LeaVerou:
We should also add this plugin on existing Mavos on our sites where we want to accept PRs, e.g. plugins.
Also, we should edit the docs about storage, if they say that PRs are enabled automatically.

DmitrySharabin:
Sure. I'll look carefully at where to make changes.

PR #634, opened 2020-08-28
Add more granular control for the autosave feature

DmitrySharabin:
Now we and plugin authors can disable autosave when needed.
E.g., when working with real-time data updates,
we don't want data loss from race condition
with autosave and realtime updates.


LeaVerou:
Thanks, merged!

However, since I imagine many plugins might want to push data, we should offer a Mavo method to do so. Plugin authors shouldn't have to think about this.

DmitrySharabin:
Yes, that would be great!

Thanks for merging.

Issue #644, opened 2020-10-19
Remove ?login after logout


leph83:
Current state: When you are logged in and press logout, the URL still has the ?login parameter, so that immediatly after you've logged out, the Popup to login shows up again. 

Remove the **?login** parameter from the URL after logout, so that the popup doesn't show up again when you are finished with editing would be a nice feature.

LeaVerou:
This should be very easy to fix.

Notes for anyone who wants to pick it up:
Basically, once someone logs in via a `?login` (and friends) params, use the history API to remove the param in place so that future link shares / reloads don't include it.

raytran:
I can take this (don't have permissions to self assign)

Issue #224, opened 2017-06-23
Login popup appears when clicking on Logout in edge case


LeaVerou:
To reproduce:

- Add `?login` in the URL without any popup blocker settings
- Login popup will be blocked.
- Then, log in normally
- Then, try to log out. The login popup appears.

PR #647, opened 2020-10-29
Fix for #644 ?login bug

raytran:
#644 #224 
Simple fix using history API to remove ?login in place before authenticating

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -177,7 +177,13 @@ var _ = self.Mavo = $.Class({
 
 		this.permissions.can("login", () => {
 			// We also support a URL param to trigger login, in case the user doesn't want visible login UI
-			if (Mavo.Functions.url("login") !== null && this.index == 1 || Mavo.Functions.url(this.id + "-login") !== null) {
+			if (Mavo.Functions.url("login") !== null && this.index === 1) {
+				// Remove ?login from url
+                const newUrl = Mavo.Functions.url().replace("?login", "");
+				window.history.pushState("", "", newUrl);
+				this.primaryBackend.login();
+			}
+			else if (Mavo.Functions.url(this.id + "-login") !== null) {
 				this.primaryBackend.login();
 			}
 		});
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -177,7 +177,7 @@ var _ = self.Mavo = $.Class({
 
 		this.permissions.can("login", () => {
 			// We also support a URL param to trigger login, in case the user doesn't want visible login UI
-            let loginUrlParam = null;
+			let loginUrlParam = null;
 			if (Mavo.Functions.url("login") !== null && this.index === 1) {
 			    loginUrlParam = "login";
 			}
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -189,7 +189,7 @@ var _ = self.Mavo = $.Class({
 				// Remove param from url
 				const currentURL = new URL(location.href);
 				currentURL.searchParams.delete(loginUrlParam);
-				history.replaceState(null, "", currentURL.toString());
+				history.replaceState(null, "", currentURL);
 				this.primaryBackend.login();
 			}
 		});
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -177,14 +177,15 @@ var _ = self.Mavo = $.Class({
 
 		this.permissions.can("login", () => {
 			// We also support a URL param to trigger login, in case the user doesn't want visible login UI
-			let loginUrlParam = null;
+			let loginUrlParam;
 			if (Mavo.Functions.url("login") !== null && this.index === 1) {
 				loginUrlParam = "login";
 			}
 			else if (Mavo.Functions.url(this.id + "-login") !== null) {
 				loginUrlParam = this.id + "-login";
 			}
-			if (loginUrlParam !== null) {
+
+			if (loginUrlParam !== undefined) {
 				// Remove param from url
 				const currentURL = new URL(location.href);
 				currentURL.searchParams.delete(loginUrlParam);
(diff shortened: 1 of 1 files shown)


src/mavo.js
```diff
@@ -177,7 +177,13 @@ var _ = self.Mavo = $.Class({
 
 		this.permissions.can("login", () => {
 			// We also support a URL param to trigger login, in case the user doesn't want visible login UI
-			if (Mavo.Functions.url("login") !== null && this.index == 1 || Mavo.Functions.url(this.id + "-login") !== null) {
+			if (Mavo.Functions.url("login") !== null && this.index === 1) {
+				// Remove ?login from url
+                const newUrl = Mavo.Functions.url().replace("?login", "");
+				window.history.pushState("", "", newUrl);
+				this.primaryBackend.login();
```
LeaVerou, line 184: So you are removing the URL param before login actually happens. Is that a conscious decision? If login fails, do we want the URL param there or not?

LeaVerou (changes requested):
Thanks for the PR Raymond! I left a bunch of comments.

LeaVerou:
Oh, and another small thing: You mention the two issues that this PR fixes in your post, but not in your commit message. However, if you mention them in your commit message with something like `fixes [issuenumber]` etc, once the PR is merged the issues will automatically be closed. See https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references

src/mavo.js
raytran: Yes this was a conscious decision on my part. At the moment if I navigate to a Mavo app with a ?login parameter and ignore the login popup, Mavo doesn't ask me to log in again. I figured it made more sense to ask once to login and redirect to the main site regardless of the login success/failure.

src/mavo.js
```diff
@@ -177,7 +177,14 @@ var _ = self.Mavo = $.Class({
...
 			// We also support a URL param to trigger login, in case the user doesn't want visible login UI
-			if (Mavo.Functions.url("login") !== null && this.index == 1 || Mavo.Functions.url(this.id + "-login") !== null) {
+			if (Mavo.Functions.url("login") !== null && this.index === 1) {
+				// Remove ?login from url
+				const currentURL = new URL(location.href);
+				currentURL.searchParams.delete("login");
+				history.replaceState(null, "", currentURL.toString());
+				this.primaryBackend.login();
+			}
+			else if (Mavo.Functions.url(this.id + "-login") !== null) {
```
LeaVerou, line 187: Why not do the same with this param? It's the same thing, it's just that instead of `login`, it's prepended with the app name, to target a specific Mavo app on the page.

LeaVerou (changes requested):
Only one issue remains!

src/mavo.js
raytran: I see; didn't realize what you meant by "There is also a more specific one, for cases when there are multiple Mavo apps on the page." earlier but this makes sense now

src/mavo.js
```diff
@@ -177,7 +177,18 @@ var _ = self.Mavo = $.Class({
 
 		this.permissions.can("login", () => {
 			// We also support a URL param to trigger login, in case the user doesn't want visible login UI
-			if (Mavo.Functions.url("login") !== null && this.index == 1 || Mavo.Functions.url(this.id + "-login") !== null) {
+			let loginUrlParam = null;
```
LeaVerou, line 180: No need to initialize to `null`, JS variables are automatically initialized to `undefined`, so you can just compare with `undefined` later.

src/mavo.js
```diff
@@ -177,7 +177,18 @@ var _ = self.Mavo = $.Class({
...
+			if (Mavo.Functions.url("login") !== null && this.index === 1) {
+				loginUrlParam = "login";
+			}
+			else if (Mavo.Functions.url(this.id + "-login") !== null) {
+				loginUrlParam = this.id + "-login";
+			}
+			if (loginUrlParam !== null) {
```
LeaVerou, line 187: Did eslint not complain about the lack of empty line before the if block? 
Currently it looks like it's part of the previous conditional. An empty line before and after control flow blocks helps visual grouping.

src/mavo.js
```diff
@@ -177,7 +177,18 @@ var _ = self.Mavo = $.Class({
...
+				// Remove param from url
+				const currentURL = new URL(location.href);
+				currentURL.searchParams.delete(loginUrlParam);
+				history.replaceState(null, "", currentURL.toString());
```
LeaVerou, line 191: No need to call `toString()`, the argument is cast to a string anyway.

LeaVerou (changes requested):
A couple extremely minor things and it's good to go!

src/mavo.js
raytran: Did not get an eslint warning on this

src/mavo.js
raytran: I actually get a warning from my IDE that "Argument type URL is not assignable to parameter type string | null | undefined   Type URL is not assignable to type string"

LeaVerou: Ok, we can leave it as-is for now. What is your IDE?

src/mavo.js
LeaVerou: Are you sure ESLint is running? [The rule is there](https://github.com/mavoweb/mavo/blob/master/.eslintrc.json#L31)

src/mavo.js
raytran: This is the checkstyle from IntelliJ Ultimate Edition

LeaVerou: Well, it appears that is a bug with your IDE. Check this out: https://github.com/whatwg/html/issues/6115

LeaVerou (approved):

LeaVerou:
Thanks, merged! Do you want to do the honours of closing the issues it fixes? (since it appears they were not closed automatically, possibly because the issues were not mentioned in the first commit? Not sure)

raytran:
> Thanks, merged! Do you want to do the honours of closing the issues it fixes? (since it appears they were not closed automatically, possibly because the issues were not mentioned in the first commit? Not sure)

Don't think I have the [permissions](https://stackoverflow.com/questions/34010565/how-do-i-close-a-github-issue-without-making-a-commit) to do that

LeaVerou:
You should be able to now!

PR #653, opened 2020-11-18
Updated duration() to include terms parameter

parulsingh23:
- Updated duration() to include a second terms parameter for number of time-units up until the user wants their value returned to be in (discarding of in-between zero-valued time-units)
- Added _.toMs in order to convert a value in terms of a given time-unit back to the value in terms of milliseconds

### DIFF
--- a/src/functions.date.js
+++ b/src/functions.date.js
@@ -117,36 +117,22 @@ _.duration = $.extend(function (ms, terms) {
 	let ret = [];
 	let units = Object.keys(s).reverse();
 	units.push("ms");
-	let unitsUsed = [];
-	let onlyConsecutiveNonzeros = true;
+	let noUnitSkipped = true;
 
-	units.map(function(unit) {
+	units.forEach(function(unit, index) {
 		// get largest value of time unit for the remaining
 		// time to account for
 		let unitValue = unit === "ms"? timeLeft : _.msTo(unit, timeLeft);
 		timeLeft -= unitValue * (unit === "ms" ? 1 : Mavo.Functions[unit]());
-		return unitValue;
-	}).forEach(function (currentValue, index) {
-		if (currentValue !== 0 && ret.length < terms) {
-			let unitProperPlurality = currentValue === 1 && units[index] !== "ms" ? units[index].slice(0, -1) : units[index];
-			ret.push(currentValue + " " + Mavo.Functions.phrase($this, unitProperPlurality));
 
-			// so we have a list of unit terms used to later determine whether they're consecutive
-			unitsUsed.push(units[index]);
-		};
-	});
-
-	ret.filter(function (currentValue, index) {
-		// terms must be consecutive
-		if (index !== 0) {
-
-			if ((units.indexOf(unitsUsed[index]) - units.indexOf(unitsUsed[index-1])) !== 1) {
-				// we will not populate the returning array further since
-				// the following units won't be consecutive
-				onlyConsecutiveNonzeros = false;
-			}
+		if (unitValue !== 0 && ret.length < terms && noUnitSkipped) {
+			let unitProperPlurality = unitValue === 1 && units[index] !== "ms" ? units[index].slice(0, -1) : units[index];
+			ret.push(unitValue + " " + Mavo.Functions.phrase(this, unitProperPlurality));
 		}
-		return onlyConsecutiveNonzeros;
+		else if (ret.length > 0) {
+			// terms are no longer consecutive terms
+			noUnitSkipped = false;
+		}
 	});
 
 	if (ret.length === 0) {
--- a/src/functions.date.js
+++ b/src/functions.date.js
@@ -114,7 +114,7 @@ _.duration = $.extend(function ($this, ms, terms) {
     }
 
     if (ms===0 || terms === undefined) {
-			terms = 1;
+terms = 1;
     }
 
     let timeLeft = ms || 0;
@@ -146,8 +146,7 @@ _.duration = $.extend(function ($this, ms, terms) {
       }
 
       if (unitTime!=0 || terms === 1) {
-        let unitProperPlurality = unitTime === 1 && unit !== "ms" ?
-					unit.slice(0, -1) : unit;
+        let unitProperPlurality = unitTime === 1 && unit !== "ms" ? unit.slice(0, -1) : unit;
         ret.push(unitTime + " " + _.phrase($this, unitProperPlurality));
         timeLeft -= unitTime * (unit === "ms" ? 1 : Mavo.Functions[unit]());
       }
@@ -158,8 +157,7 @@ _.duration = $.extend(function ($this, ms, terms) {
       }
     }
 
-    return (arguments.length ===1 ||
-			(arguments.length ===2 && $this!== undefined))? ret[0] : ret;
+    return (arguments.length ===1 || (arguments.length ===2 && $this!== undefined))? ret[0] : ret;
 }, {
     needsContext: true
 });
--- a/src/functions.date.js
+++ b/src/functions.date.js
@@ -109,57 +109,57 @@ for (let unit in s) {
 }
 
 _.duration = $.extend(function ($this, ms, terms) {
-    if (arguments.length === 1) {
-      [ms, $this] = [$this, null];
-    }
+		if (arguments.length === 1) {
+			[ms, $this] = [$this, null];
+		}
 
-    if (ms===0 || terms === undefined) {
-      terms = 1;
-    }
+		if (ms===0 || terms === undefined) {
+			terms = 1;
+		}
 
-    let timeLeft = ms || 0;
-    let ret = [];
-    let unitsToUse = ["ms"];
+		let timeLeft = ms || 0;
+		let ret = [];
+		let unitsToUse = ["ms"];
 
-    // populate unitsToUse with all the terms that can fit within timeLeft
-    // in order of lowest to highest magnitude,
+		// populate unitsToUse with all the terms that can fit within timeLeft
+		// in order of lowest to highest magnitude,
 		// so we have a list of consecutive unit terms to later reference
-    for (let nextUnit in s) {
-      let count = _.msTo(nextUnit, timeLeft);
+		for (let nextUnit in s) {
+			let count = _.msTo(nextUnit, timeLeft);
 
-      if (count === 0) {
-        break;
-      }
+			if (count === 0) {
+				break;
+			}
 
-      unitsToUse.push(nextUnit);
-    }
+			unitsToUse.push(nextUnit);
+		}
 
-    while (ret.length < terms && unitsToUse.length > 0) {
-      let unit = unitsToUse.pop();
+		while (ret.length < terms && unitsToUse.length > 0) {
+			let unit = unitsToUse.pop();
 			let unitTime;
 
-      if (unit!=="ms") {
-        unitTime = _.msTo(unit, timeLeft);
-      }
-      else {
-        unitTime = timeLeft;
-      }
+			if (unit!=="ms") {
+				unitTime = _.msTo(unit, timeLeft);
+			}
+			else {
+				unitTime = timeLeft;
+			}
 
-      if (unitTime!=0 || terms === 1) {
-        let unitProperPlurality = unitTime === 1 && unit !== "ms" ? unit.slice(0, -1) : unit;
-        ret.push(unitTime + " " + _.phrase($this, unitProperPlurality));
-        timeLeft -= unitTime * (unit === "ms" ? 1 : Mavo.Functions[unit]());
-      }
-      else {
-        // we will not populate the returning array further since
-        // the following units won't be consecutive
-        break;
-      }
-    }
+			if (unitTime!=0 || terms === 1) {
+				let unitProperPlurality = unitTime === 1 && unit !== "ms" ? unit.slice(0, -1) : unit;
+				ret.push(unitTime + " " + _.phrase($this, unitProperPlurality));
+				timeLeft -= unitTime * (unit === "ms" ? 1 : Mavo.Functions[unit]());
+			}
+			else {
+				// we will not populate the returning array further since
+				// the following units won't be consecutive
+				break;
+			}
+		}
 
-    return (arguments.length ===1 || (arguments.length ===2 && $this!== undefined))? ret[0] : ret;
+		return (arguments.length ===1 || (arguments.length ===2 && $this!== undefined))? ret[0] : ret;
 }, {
-    needsContext: true
+		needsContext: true
 });
 
 $.extend(_.util, {
--- a/src/functions.date.js
+++ b/src/functions.date.js
@@ -114,7 +114,7 @@ _.duration = $.extend(function ($this, ms, terms) {
     }
 
     if (ms===0 || terms === undefined) {
-terms = 1;
+ terms = 1;
     }
 
     let timeLeft = ms || 0;
@@ -128,7 +128,7 @@ terms = 1;
       let count = _.msTo(nextUnit, timeLeft);
 
       if (count === 0) {
-				break;
+   break;
       }
 
       unitsToUse.push(nextUnit);
--- a/src/functions.date.js
+++ b/src/functions.date.js
@@ -108,27 +108,32 @@ for (let unit in s) {
 	}, {multiValued: true});
 }
 
-_.duration = $.extend(function($this, ms, terms = 1) {
+_.duration = $.extend(function($this, ms, terms) {
 	if (arguments.length === 1) {
 		[ms, $this] = [$this, null];
 	}
 	
+	let wantString = terms === undefined ? true : false;
 	let unitTime = ms || 0;
 	let timeLeft = ms || 0;
 	
-	if (ms === 0) {
+	if (ms === 0 || wantString) {
 		terms = 1;
 	}
 	
-	let timeArray = [];
+	let ret = [];
 	
-	while (timeArray.length < terms && (timeLeft > 0 || terms === 1)) {
+	while (ret.length < terms && (timeLeft > 0 || terms === 1)) {
+		// Only add terms if more than 1 term was designated, and those terms must be nonzero,
+      		// else, if one term was desitnated, that term can be nonzero
 		let unit = "ms";
 		
 		for (let nextUnit in s) {
 			let count = _.msTo(nextUnit, timeLeft);
 			
 			if (count === 0) {
+				// next higher-magnitude term occurs 0 times within timeLeft,
+          			// so remain with last loop's assigned unit & unitTime
 				break;
 			}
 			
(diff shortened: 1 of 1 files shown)


parulsingh23:
The duration function now takes a second argument specifying the number of terms you want the returned string up until as a second argument (default is 1).
Given the number of terms, you output the time units with a nonzero value (discarding the time units with a 0 value as they don’t count as a term - only exception is for duration(0) which outputs a single term of ‘0 ms’ regardless of the terms argument specified). For example, calling duration(60*60*1000+1000, 2) returns ‘1 hour 1 second’ (for any terms argument of 2 or greater).

src/functions.date.js
```diff
@@ -98,6 +98,16 @@ $.extend(_, {
 
 _.msTo = (what, ms) => Math.floor(Math.abs(ms) / (s[what] * 1000)) || 0;
 
+_.toMs = function (what, value) {
```
LeaVerou, line 101: If I understand what this does correctly, I don't think it's needed. We typically do it via e.g. `5 * days()` in Mavo expressions, which translates to `5 * Mavo.Functions.days()` in JS.

LeaVerou (changes requested):
Hi Parul, thanks for submitting this PR!

Sme low hanging fruit comments of things I spotted from a first skim:

* Returning the terms like "1 hour 2 minutes" reduces utility because you cannot display that anywhere. I would suggest returning an array, which if printed as-is gets joined with ", " (so it would read "1 hour, 2 minutes"), but it can also be used in `mv-value` for more complex use cases.
* Please install ESLint, including an ESLint plugin for your editor to ensure you follow the coding style of the project. This is important to ensure consistency in a codebase that multiple people contribute to. There are several violations here (e.g. indentation, spacing around operators, spacing before parens etc).
* Prefer `let` over `var` which is block scoped. There are certain bugs you can run into with `var` that `let` would prevent.
* Use blank lines to separate control flow statements, clean up blank lines that don't have this purpose.

Also see inline comments.

Let's fix those, and I can review the algorithm properly in the next iteration. :)

parulsingh23:
I updated the duration function to now return an array and removed the unnecessary toMs() function. I have my ESLinter plugin running on atom, however, it gave no indication on blank lines (so I just added them between the control statements as you mentioned) and also missed some extra spaces I had (which I later went in and removed). But let me know if there's any other things I should address, thanks!

src/functions.date.js
```diff
@@ -108,28 +108,43 @@ for (let unit in s) {
...
+			unit = nextUnit;
 		}
-
-		count = nextCount;
-		unit = nextUnit;
+		
+		if (unitTime !== 0 || terms === 1) {
+			let unitProperPlurality = unitTime === 1 && unit !== "ms" ? unit.slice(0, -1) : unit;
+			timeArray.push(unitTime + " " + _.phrase($this, unitProperPlurality));
+			timeLeft -= unit === "ms" ? unitTime : unitTime * Mavo.Functions[unit]();
```
LeaVerou, line 142: I would put the ternary inside the second operand:

```js
timeLeft -= unitTime * (unit === "ms" ? 1 : Mavo.Functions[unit]());
```

src/functions.date.js
```diff
@@ -108,28 +108,43 @@ for (let unit in s) {
...
+		
+		for (let nextUnit in s) {
+			let count = _.msTo(nextUnit, timeLeft);
+			
+			if (count === 0) {
+				break;
+			}
+			
+			unitTime = count;
+			unit = nextUnit;
```
LeaVerou, line 136: I may have misunderstood your algorithm, but it seems to me that this inner loop only runs once if `_.msTo()` returns 0, otherwise it keeps going and overwriting previous results? Why is that?

In general, this is at the point of complexity that it would have benefitted from a few comments, at least to explain what this loop does, and what the conditional below (the rest of the codebase is not necessarily a shining example of sufficient comments usage either, I know…).

src/functions.date.js
```diff
@@ -108,28 +108,43 @@ for (let unit in s) {
...
-			break;
+	
+	let unitTime = ms || 0;
+	let timeLeft = ms || 0;
+	
+	if (ms === 0) {
+		terms = 1;
+	}
+	
+	let timeArray = [];
```
LeaVerou, line 123: While in general abbreviations are a bad idea, there is the widespread convention of calling the value that will be eventually returned `ret`, which means that someone reading the code can instantly tell that this is what will eventually be returned, which aids understanding. You may or may not want to follow it. The rest of the codebase mostly does.

src/functions.date.js
```diff
@@ -108,28 +108,43 @@ for (let unit in s) {
...
+	
+	while (timeArray.length < terms && (timeLeft > 0 || terms === 1)) {
```
LeaVerou, line 125: Why `terms === 1` as a separate condition? Wouldn't the rest suffice even in that case?

LeaVerou (changes requested):
Much better now! I added some minor points and a question. 

Also, you should add some tests to the testsuite. I see we have none for `duration()`, so you'd need to add some for basic usage too: https://test.mavo.io/functions.html

LeaVerou:
Also, I just realized something:

When `duration()` is called without a `terms` parameter, we want to return a string. Only when it's used with an explicit `terms` parameter we want to return an array.

This means that `duration(foo, 1)` is different than `duration(foo)`, so we can't use a default value for `terms` as we need to distinguish whether it was provided or not (parameters not provided have the value `undefined`)

src/functions.date.js
```diff
@@ -108,28 +108,43 @@ for (let unit in s) {
 	}, {multiValued: true});
 }
 
-_.duration = $.extend(function($this, ms) {
+_.duration = $.extend(function($this, ms, terms = 1) {
 	if (arguments.length === 1) {
 		[ms, $this] = [$this, null];
```
LeaVerou, line 113: Also, another thing I forgot in my previous review: You added a parameter, but did not change this part of the code (which essentially deals with `$this` not being provided).

src/functions.date.js
parulsingh23: As long as we have not filled up our array with the allotted # of terms, we will enter our while loop to add more terms if there is timeLeft to account for, or we had only 1 term asked for (in which case it's ok to return a zero value). So it allows for zero terms to be returned only in that zero term is the only term being returned. I'll add this in a comment to the code.

src/functions.date.js
parulsingh23: Good point; I'll add in a clarifying comment. This loop only runs once if _.msTo() returns 0 b/c nextUnit starts with the lowest-valued term apart from ms: seconds. So if a seconds conversion of the timeLeft is 0, then we will stick with the current unit of ms (which is why we break).

src/functions.date.js
parulsingh23: I am a bit confused as to how we know whether $this has been provided if we have 2 arguments. Would it be that the type of $this is a number when it should be an object?

parulsingh23:
I have updated duration() to include the comments (but definitely let me know if you think I should add more explanation for them), and added duration tests. 
However, I am not yet sure how we deal with the scenario of identifying whether or not $this has been provided if we are passed in 2 arguments.

src/functions.date.js
```diff
@@ -108,27 +108,32 @@ for (let unit in s) {
 	}, {multiValued: true});
 }
 
-_.duration = $.extend(function($this, ms, terms = 1) {
+_.duration = $.extend(function($this, ms, terms) {
 	if (arguments.length === 1) {
 		[ms, $this] = [$this, null];
 	}
 	
+	let wantString = terms === undefined ? true : false;
```
LeaVerou, line 116: `x? true : false` is an antipattern: what you're essentially doing is converting `x` to boolean, so you can express it explicitly as `Boolean(x)` (or `!!x`, though that's less readable). In this case `terms === undefined` is *already* a boolean, so you can just do `let wantString = terms === undefined;`. 

Also, I don't think it's a good idea to have the logic about what to return here, I think it would be easier to read closer to the actual return statement. You can also use `arguments.length` to avoid having to store a variable about whether to return a string or not.

So it can be as succinct as `return arguments.length < 2? ret[0] : ret;` (though it would need to account for no `$this` being provided so it will be a little more complex than that).

src/functions.date.js
```diff
@@ -138,13 +143,18 @@ _.duration = $.extend(function($this, ms, terms = 1) {
...
 		}
 		
 	}
-	return timeArray;
+	
+	if (wantString) {
+      		return ret[0];
+    	}
+		
+    	return ret;
```
LeaVerou, line 157: Rogue spaces for indentation. Spaces are ok for alignment, but we use tabs for indentation (this is called "smart tabs"). I thought ESLint warned about that?

src/functions.date.js
```diff
@@ -108,27 +108,32 @@ for (let unit in s) {
...
+		// Only add terms if more than 1 term was designated, and those terms must be nonzero,
+      		// else, if one term was desitnated, that term can be nonzero
 		let unit = "ms";
 		
 		for (let nextUnit in s) {
 			let count = _.msTo(nextUnit, timeLeft);
 			
 			if (count === 0) {
+				// next higher-magnitude term occurs 0 times within timeLeft,
+          			// so remain with last loop's assigned unit & unitTime
```
LeaVerou, line 136: I don’t find either of these comments helpful in understanding the code. They seem to be describing **what** the code is doing, but good comments describe the **why**. The what should be self-evident from the code, it's the **why** that we need additional context for.
See https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

LeaVerou (changes requested):

src/functions.date.js
LeaVerou: It's not that simple, because `ms` can also come from a property, in which case they would be indistinguishable. Let me think about it and get back to you.

parulsingh23:
I updated duration() to only include consecutive nonzero terms. However, I was unsure about the last returning line of code: is it correct for (arguments.length ===2 && $this!== undefined) to mean that the terms parameter hasn't been given and that we're only returning a single term?

LeaVerou:
Hi Parul, I see you're committing. Please don't forget to tell me when you're ready for me to review again.
I noticed a lot of commits with incorrect indentation etc. As a general practice, it helps catch a lot of issues to preview what is about to be committed, either on the Github Desktop app, or via `git diff` in the console. This also ensures a more solid commit history, without too minor commits that are created to rectify issues that previous commits introduced because they were done blindly.
For indentation specifically, it also helps to enable "Show invisibles" in your editor which displays tabs and spaces differently, e.g.:
image

parulsingh23:
@LeaVerou you can review now! And thanks for the pointer on enabling "Show invisibles" in the editor. That made it a lot clearer why I was having indentation alignment in the editor but not on the GitHub files (I had a mix of lines indented with spaces & others with tabs, so I shifted it to be indented by tabs as the rest of the file is).

LeaVerou:
Hi Parul,

Indeed, it's now correctly indented with tabs, but it looks like everything is indented one level more than it needs to?

parulsingh23:
I see, fixed it!

src/functions.date.js
```diff
@@ -108,28 +108,56 @@ for (let unit in s) {
...
+	let timeLeft = ms || 0;
+	let ret = [];
+	let unitsToUse = ["ms"];
 
+	// populate unitsToUse with all the terms that can fit within timeLeft
+	// in order of lowest to highest magnitude,
+	// so we have a list of consecutive unit terms to later reference
 	for (let nextUnit in s) {
-		var nextCount = _.msTo(nextUnit, ms);
+		let count = _.msTo(nextUnit, timeLeft);
```
LeaVerou, line 128: I think there's a bug here. `timeLeft` sounds like it would be updated in every iteration but it's not, so essentially you are always converting `ms`?

Since you are always converting the same number, you may have something that converts to an integer number for a given unit but once that is subtracted, it would be 0.
E.g. `_.msTo("days", 24 * 60 * 60 * 1000)` returns 1, but `_.msTo("hours", 24 * 60 * 60 * 1000)` returns 24, so it would be included.

I think you don't need two loops, you can just get all candidate units via `Object.keys(s)` and loop over that.

Also, I'd call that just `units`. Of course they are "to use", otherwise why are we defining them?

src/functions.date.js
```diff
@@ -108,28 +108,56 @@ for (let unit in s) {
...
+	while (ret.length < terms && unitsToUse.length > 0) {
+		let unit = unitsToUse.pop();
+		let unitTime;
+
+		if (unit!=="ms") {
+			unitTime = _.msTo(unit, timeLeft);
+		}
+		else {
+			unitTime = timeLeft;
+		}
```
LeaVerou, line 146: Maybe a ternary here?

`let unitTime = unit === "ms"? timeLeft : _.msTo(unit, timeLeft);`

Also, I'm not sure `unitTime` is very descriptive, I had to follow the code to understand what it's supposed to be. However I don't have any suggestions for a better name right now.

src/functions.date.js
```diff
@@ -108,28 +108,56 @@ for (let unit in s) {
...
+
+		if (unitTime!=0 || terms === 1) {
+			let unitProperPlurality = unitTime === 1 && unit !== "ms" ? unit.slice(0, -1) : unit;
+			ret.push(unitTime + " " + _.phrase($this, unitProperPlurality));
+			timeLeft -= unitTime * (unit === "ms" ? 1 : Mavo.Functions[unit]());
```
LeaVerou, line 151: Isn't `_` equal to `Mavo.Functions` here?

src/functions.date.js
```diff
@@ -108,28 +108,56 @@ for (let unit in s) {
...
+		}
+		else {
+			// we will not populate the returning array further since
+			// the following units won't be consecutive
+			break;
+		}
+	}
 
-	return count + " " + _.phrase($this, unit);
+	return (arguments.length ===1 || (arguments.length ===2 && $this!== undefined))? ret[0] : ret;
```
LeaVerou, line 160: Perhaps a variable for the condition here to make it more clear what we're checking?

LeaVerou (changes requested):
Sorry for the delay Parul, see comments!

parulsingh23:
@LeaVerou you can take a look at the updates I made and let me know of any improvements I should make, thanks!

src/functions.date.js
```diff
@@ -108,24 +108,52 @@ for (let unit in s) {
...
+	units.map(function(unit) {
+		// get largest value of time unit for the remaining
+		// time to account for
+		let unitValue = unit === "ms"? timeLeft : _.msTo(unit, timeLeft);
+		timeLeft -= unitValue * (unit === "ms" ? 1 : Mavo.Functions[unit]());
+		return unitValue;
+	}).forEach(function (currentValue, index) {
+		if (currentValue !== 0 && ret.length < terms) {
+			let unitProperPlurality = currentValue === 1 && units[index] !== "ms" ? units[index].slice(0, -1) : units[index];
+			ret.push(currentValue + " " + Mavo.Functions.phrase($this, unitProperPlurality));
```
LeaVerou, line 132: Parul, did this code run? `$this` will throw an error here, check out the code you're modifying: it includes `this`, not `$this`. `$this` is only a thing within Mavo expressions.

src/functions.date.js
```diff
@@ -108,24 +108,52 @@ for (let unit in s) {
...
+	let unitsUsed = [];
+	let onlyConsecutiveNonzeros = true;
+
+	units.map(function(unit) {
+		// get largest value of time unit for the remaining
+		// time to account for
+		let unitValue = unit === "ms"? timeLeft : _.msTo(unit, timeLeft);
+		timeLeft -= unitValue * (unit === "ms" ? 1 : Mavo.Functions[unit]());
+		return unitValue;
+	}).forEach(function (currentValue, index) {
```
LeaVerou, line 129: If you're not using `map()`'s return value, what's the point of the subsequent `forEach()`? Why not use a single looping construct (whether that's `map()`, `forEach()`, `for .. of` loop or something else)?

src/functions.date.js
```diff
@@ -108,24 +108,52 @@ for (let unit in s) {
...
+		if (currentValue !== 0 && ret.length < terms) {
+			let unitProperPlurality = currentValue === 1 && units[index] !== "ms" ? units[index].slice(0, -1) : units[index];
+			ret.push(currentValue + " " + Mavo.Functions.phrase($this, unitProperPlurality));
+
+			// so we have a list of unit terms used to later determine whether they're consecutive
+			unitsUsed.push(units[index]);
+		};
+	});
+
+	ret.filter(function (currentValue, index) {
```
LeaVerou, line 139: This seems a little overengineered, unless I'm missing some complexity of the check. It seems to me that if we encounter a zero, and it's not the first term, we just stop there and don't calculate any further terms, right? If so, it should be possible to just have a single loop (possibly even a simple `for` loop with an index), and just `break;` if `i > 0 && value === 0`. Am I missing something?

LeaVerou (changes requested):
Hi Parul, see comments!

src/functions.date.js
parulsingh23: I'm running the duration code in the mavo.es5.js file, which doesn't seem to incorporate the removal of the `$this` param for the function, so I had just assumed for the src file I'd manually take that out. But I forgot to remove it from this phrase function call, thanks for the catch! Is there a way to have the mavo.es5.js file run the same duration version (without `$this`) as the src does?

parulsingh23:
@LeaVerou I updated the duration function, so you can take a look, thanks! I now have a single forEach loop running on `units` and the boolean`noUnitSkipped` set to false if the nonZero consecutive terms streak is 'broken' and the ret array is then no longer updated. I don't believe we can break out of a forEach loop, which is why I am using the boolean to signal no further populating of the ret array. But definitely let me know if you think another approach is better!

src/functions.date.js
LeaVerou: Wait, you should never be manually modifying files in `dist/`, those should be generated by gulp. You only modify files in `src`.
`$this` previously was a function argument. If it's removed as a function argument, it would throw. Which bundle you're running off shouldn't have anything to do with it unless something weird is going on.

LeaVerou:
If a looping function doesn't work for your use case, don't use a looping function. If you need to `break` out of the loop, a classic `for(let i=0; i<n; i++)` is perfectly fine. Traditional `for` loops have their place too, they have not become passé because we now have looping functions. The right tool for the right job 😄 
Regardless, since this works, I'm going to merge this and make a few edits myself (and CC you in the commit), since I don't want us to go to winter break before merging this PR. :)

LeaVerou:
@parulsingh23 I just pushed a commit to fix an error thrown with the call to `phrase()`, see https://github.com/mavoweb/mavo/commit/c617d8d8de3f2b8ab062f4dcf439dc541bf566b1

The function was throwing, because it was getting unexpected arguments. Please take a look at commit d26e7020d8c77790dbeb8dcc1b83b030db0f0957 and compare the `phrase()` call that was replaced with the `phrase()` call that replaced it. Note the missing `.call()`, which would pass `this` as context instead of first argument.

I know it's often very tempting to make small commits and push them without testing if the code runs (I've done it myself many times!) but it's easy to overlook something small like that and computers are unforgiving (making them more forgiving is actually a research direction).

PR #654, opened 2020-11-19
Fixes #640

### LINKED ISSUE #640: If an expression contains a mistake, Mavo shows `[SyntaxError: Unexpected token ')']` as its result
Previously Mavo showed the expression itself (as it should), but now something is broken.
Testcase: https://mavo.io/docs/expressions#providing-fallbacks-the-mv-value-attribute

DmitrySharabin:
We need to save the source expression to use it as a fallback and show it to a user as it is described in the docs: https://mavo.io/docs/expressions#providing-fallbacks-the-mv-value-attribute.

### DIFF
--- a/src/expression.js
+++ b/src/expression.js
@@ -59,7 +59,9 @@ Bliss.Class(_, {
 
 				Mavo.hooks.run("expression-compile-error", {context: this, error});
 
-				return this.function = error;
+				this.function = error;
+
+				return value;
 			}
 
 			this.ast = this.options.ast;


DmitrySharabin:
I also checked that all the expressions tests of the test suite (https://test.mavo.io/expressions.html) are passed.

LeaVerou:
Thank you for working on this. I hastily merged it, but then I remembered that there is actually code in DOMExpression that depends on Error return values: https://github.com/mavoweb/mavo/blob/master/src/domexpression.js#L153 so I'm gonna revert it for now.

LeaVerou:
Although, that seems to be around runtime errors, not syntax errors, so maybe it's ok.

LeaVerou:
Though, it is questionable whether it's actually a good idea to not provide any feedback to the user that their expression contains a syntax error.

DmitrySharabin:
> Thank you for working on this. I hastily merged it, but then I remembered that there is actually code in DOMExpression that depends on Error return values: https://github.com/mavoweb/mavo/blob/master/src/domexpression.js#L153, so I'm gonna revert it for now.

Yes, but `env.value` is a result of an expression evaluation: https://github.com/mavoweb/mavo/blob/243feed58e3a57743262bd8234e63bf9d05fbe4e/src/domexpression.js#L147

Since at the moment of the expression creation we get a compile error, the `this.function` contains this error, the `eval` method will return it, so the `env.value` will get the exact same compile error as a result. And everything will go as It should. I worked through this case. And that was a tough investigation. :)

DmitrySharabin:
> Though, it is questionable whether it's actually a good idea to not provide any feedback to the user that their expression contains a syntax error.

Agree. Especially if a user doesn't know about developer tools.

I only tried to recover the documented functionality.

DmitrySharabin:
One more thought. The error messages we get from JSEP are not very helpful. Not only for Mavo authors but also for us. E.g., for `[42 +]` we get `Unexpected token ')'` even though there is not a single `)` in the expression we are trying to parse.

LeaVerou:
Yeah, that's because the expression is wrapped in parentheses before evaluated. Perhaps if there's a syntax error we could re-evaluate the isolated expression and grab the syntax error from that. It will still be very technical, but at least it won't complain about tokens that you can't even see.

LeaVerou:
But the question of how to present these errors remains.

PR #658, opened 2020-11-30
Implement storage attributes, resolves #657

### LINKED ISSUE #657: Allow selecting/configuring backend metadata via attributes
_(In the following text, `mv-storage` refers to all four backend-selecting attributes (`mv-storage`, `mv-source`, `mv-init`, `mv-uploads`)_

Currently, both selecting and configuring a backend is done entirely via the `mv-storage` value, which is all that's passed from Mavo to the backend object. This has several drawbacks:

* For some backends, a URL simply doesn't make sense, so they are forced to be configured via nonsensical, nonexistent URLs (e.g. [Firebase](https://plugins.mavo.io/plugin/firebase-firestore#setup-mavo-application) )
* Regardless of whether a URL makes sense, many backends need additional configuration parameters (e.g. [Firebase](https://plugins.mavo.io/plugin/firebase-firestore#setup-mavo-application) but many others too).
* REST URLs have some of the same drawbacks of positional arguments in functions. E.g. with Github, how do you specify a repo name without a username?

A nice solution to these issues would be for Mavo to pick up all `mv-storage-*` attributes, package them neatly into an object literal, and use them both for matching which backend to use, and as an argument to the backend constructor. These could also be used in backends that *do* use URLs, for an alternative, more readable way, to specify the same parameters, or for a way to skip some parameters (e.g. `mv-storage="https://github.com" mv-repo="foobar"` allows specifying a a repo without a username)

LeaVerou:
* Code to read attributes prefixed with `mv-storage-*`, `mv-source-*` etc and send them to the backend code as an object literal
* Replaces ad hoc code for `mv-storage-type` and `mv-storage-format`
* I ported `mv-dropbox-key` and `mv-github-key` to use this new syntax (backwards incompatible since now they'll be `mv-storage-key` etc, but I don't think there is any usage of custom API keys). I also ported that code to `backend.js` since that's where it's actually used.
* There is no code yet to allow attributes to override Github params like username or repo since that turned out to be a little more complicated, so it's better done in a separate commit/PR.

### DIFF
--- a/src/backend.dropbox.js
+++ b/src/backend.dropbox.js
@@ -6,8 +6,6 @@ var _ = Mavo.Backend.register($.Class({
 	constructor: function() {
 		this.permissions.on(["login", "read"]);
 
-		this.key = this.mavo.element.getAttribute("mv-dropbox-key") || "2mx6061p054bpbp";
-
 		this.login(true);
 	},
 
@@ -89,6 +87,7 @@ var _ = Mavo.Backend.register($.Class({
 	static: {
 		apiDomain: "https://api.dropboxapi.com/2/",
 		oAuth: "https://www.dropbox.com/oauth2/authorize",
+		key: "2mx6061p054bpbp",
 
 		test: function(url) {
 			url = new URL(url, Mavo.base);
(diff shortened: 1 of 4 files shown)


DmitrySharabin (approved):
Thank you for letting me take a look at this PR. I'll slightly change the code of the GSheets backend to reflect the changes you made. Luckily, this PR doesn't introduce any breaking changes to the Firebase backend. However, it may help simplify it. :)

PR #669, opened 2020-12-14
Make the mv-backendType-* family of attributes live

DmitrySharabin:
It might be I wasn't supposed to take this issue since it's labeled as a good first issue. 🙈 I just thought that it would be nice if we have this feature asap.

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -161,7 +161,9 @@ var _ = self.Mavo = $.Class({
 
 		this.observe({deep: false, attribute: true}, ({attribute}) => {
 			if (attribute.indexOf("mv-") === 0) {
-				let role = attribute?.replace(/^mv-/, "");
+				// We want to observe changes both in a backend (the mv-role attribute)
+				// and its metadata (provided via the mv-role-* family of attributes)
+				let [role, ...rest] = attribute?.replace(/^mv-/, "")?.split("-");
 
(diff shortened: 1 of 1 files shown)


src/mavo.js
```diff
@@ -161,7 +161,9 @@ var _ = self.Mavo = $.Class({
 
 		this.observe({deep: false, attribute: true}, ({attribute}) => {
 			if (attribute.indexOf("mv-") === 0) {
-				let role = attribute?.replace(/^mv-/, "");
+				// We want to observe changes both in a backend (the mv-role attribute)
+				// and its metadata (provided via the mv-role-* family of attributes)
+				let [role, ...rest] = attribute?.replace(/^mv-/, "")?.split("-");
```
LeaVerou, line 166: It seems a bit misleading to use a variable that is never used. What about:

```js
let role = attribute?.replace(/^mv-/, "")?.split("-")?.[0];
```

LeaVerou (commented):
Thanks for working on this! See (minor) comment (which is only an idea, don't feel obligated to follow that)

src/mavo.js
DmitrySharabin: You are absolutely right! That's better. Fixed. Thank you. 🙏🏻

LeaVerou:
Thanks, merged!

PR #670, opened 2020-12-15
[github] Use attributes as another way to provide metadata

### LINKED ISSUE #665: [github] Allow overriding Github params via attributes
Leftover work from #657
Blocked on #664

Now that we can provide metadata to the backend via attributes, we should also support both when this makes sense. For example, for Github we should be able to override the params that are extracted from the URL through these attributes, whether they are provided in the URL or not. 
Any data not provided should default to the existing defaults.

E.g. these should be all valid:

```html
<E mv-storage="https://github.com" mv-storage-username="leaverou" mv-storage-repo="mavoweb" mv-storage-path="data.json">
<E mv-storage="https://github.com/karger/otherrepo" mv-storage-username="leaverou" mv-storage-repo="mavoweb" mv-storage-path="data.json">
<E mv-storage="https://github.com/leaverou"mv-storage-repo="mavoweb" mv-storage-path="data.json">
<E mv-storage="https://github.com" mv-storage-username="leaverou" mv-storage-path="data.json">
```

This is trickier than it seems due to how `Mavo.Backend.Github.parseURL()` is implemented, where some values depend on other values.

DmitrySharabin:
Let's try to fix #665 (or at least begin iterating over it that I find beneficial). 😀

### DIFF
--- a/src/backend.github.js
+++ b/src/backend.github.js
@@ -35,12 +35,18 @@ var _ = Mavo.Backend.register($.Class({
 		}
 
 		// Let an author provide either filepath, filename, or path.
-		// The path property takes priority.
-		// Leave raw API calls as-is for now
+		// The path property takes priority
 		if (this.info.path) {
-			this.info.path = o.path ?? (this.info.filepath ? this.info.filepath + "/" : "") + this.info.filename;
+			if (!o.path) {
+				if (this.info.filename) {
+					this.info.path = (this.info.filepath? this.info.filepath + "/" : "") + this.info.filename;
+				}
+				else {
+					this.info.path = this.info.filepath;
+				}
+			}
 
-			this.info.apiCall = `repos/${this.info.username}/${this.info.repo}/contents/${this.info.path}`;
+			this.info.apiCall = `repos/${this.info.username}/${this.info.repo}/contents/${this.info.path}` + (this.info.apiParams ?? "");
 		}
 
 		$.extend(this, this.info);
@@ -67,7 +73,7 @@ var _ = Mavo.Backend.register($.Class({
 					headers: {
 						"Accept": "application/vnd.github.squirrel-girl-preview"
 					}
-				}).then(response => Promise.resolve(info.repo? _.atob(response.content) : response));
+				}).then(response => Promise.resolve(info.repo && response.content? _.atob(response.content) : response));
 		}
 		else {
 			// Unauthenticated, use simple GET request to avoid rate limit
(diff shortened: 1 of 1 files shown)


src/backend.github.js
```diff
@@ -383,7 +383,13 @@ var _ = Mavo.Backend.register($.Class({
 				ret.filename = defaults.filename;
 			}
 
-			ret.filepath = path.join("/") || defaults.filepath || "";
+			// If an author provided backend metadata, use them
+			// since they have higher priority
+			for (const prop in ret) {
```
LeaVerou, line 388: One issue with this is that it only iterates over properties already defined on `ret`, but a common case would be to define things via attributes and NOT in the URL, so they may not exist as keys on `ret`.

LeaVerou (commented):
Hi Dmitry, thanks for working on this!
This largely works (see my one comment though).
The reason I didn't want to go that route is that it would be a better architecture if `parseURL()` just parses the URL like the function name implies, returns its components, and we override them as needed in the constructor. However, a big problem with that approach is that certain components depend on others, e.g. apiCall, path.
I'm not sure if the implementation simplicity is worth the architectural tradeoff.

src/backend.github.js
```diff
@@ -20,7 +20,7 @@ var _ = Mavo.Backend.register($.Class({
 			filename: `${this.mavo.id}${extension}`
 		};
 
-		this.info = _.parseURL(this.source, this.defaults);
+		this.info = _.parseURL(this.source, this.defaults, o);
```
LeaVerou, line 23: Do note there's another call to `parseURL` in `get()`. Do we need to modify that too?

DmitrySharabin:
Hi Lea!

Thank you for the comments. I agree with your thoughts and share them.

> I'm not sure if the implementation simplicity is worth the architectural tradeoff.

Of course, it doesn't. Well, I continue working then. :)

LeaVerou:
Btw, I did mean I wasn't sure. If anything, I was leaning towards that the tradeoff might be worth it. But if you want to look into it a bit more first, that's fine!

DmitrySharabin:
Yes, I'd like to look into it a bit more. Every single time I learn a lot from these iterations. And I appreciate that. Thank you.

DmitrySharabin:
Hi @LeaVerou,

Will you please give this PR another look when you have time? I made some adjustments and would love to have feedback from you.

Thanks!

LeaVerou:
Thank you for working on this Dmitry. I will review properly in a bit.
Thought: I wonder if it would simplify things to express computed properties like `apiCall` as a writable accessor.
What do I mean by that, since you can't both specify accessors and `writable: true`? Something like this:

```js
Object.defineProperty(ret, "apiCall", {
	get () { 
		return /* expression */
	}, 
	set (v) {
		delete this.apiCall;
		this.apiCall = v;
	},
	configurable: true, 
	enumerable: true
});
```

DmitrySharabin:
I saw a similar code in the Mavo code base, but, honestly, I’ve never used it myself. It’s not in my blood. Yet. :) Thank you for your suggestion (question). I’ll examine this object method on MDN (or elsewhere) and give it a try. I will be back soon. ;)

DmitrySharabin:
@LeaVerou Thank you so much for your suggestion and the brilliant following blog post that helped me A LOT. If I got everything correctly, the code we have now is more maintainable, DRY, and without any doubt, more professional. 🤞🏻 The last word is up to you, of course. ☺️

Your help is priceless. I appreciate that. Thank you! 🙇

LeaVerou (approved):
LGTM. Nice work, Dmitry.

I think we should move the writable getters to a helper in `util.js` (or even one in Bliss), but we can do that via a separate PR later.

Next step would be to add some tests for this, both for specifying params via attributes that *aren't* present in the URL, as well as *overriding* params that are present.

DmitrySharabin:
Thank you, Lea!

> Next step would be to add some tests for this, both for specifying params via attributes that aren't present in the URL, as well as overriding params that are present.

Yes, it's on my radar.

> I think we should move the writable getters to a helper in util.js (or even one in Bliss), but we can do that via a separate PR later.

Vote for Bliss. :) Would love to contribute to that awesome project too. ☺️

LeaVerou:
> Vote for Bliss. :) Would love to contribute to that awesome project too. ☺️

Then hold off for a bit longer, I'm rewriting bliss to use ESM so until that ships, no new features.

DmitrySharabin:
Deal! 🤝

PR #674, opened 2020-12-24
Attempt to resolve #248

### LINKED ISSUE #248: Set timestamp only on storage backends that require it
Right now, a [timestamp is set on all requests](https://github.com/mavoweb/mavo/blob/v0.1.1/src/backend.js#L18) by default. However, this might only be a requirement for certain storage backends (GitHub, for instance).

The downside of this automatic append is that it breaks caching (for stores that correctly support it, which does not include GitHub), and that it might break on backends that do not support additional query parameters (they could then override, but perhaps not adding anything might be a sensible default).

Does it make sense to:
- delegate that responsibility to the storage backends?
- or hide that functionality behind a flag?

LeaVerou:
This is one way to resolve it. Another way would be for backends to define the option as a static property, in the same way they define their OAuth domain etc.

Thoughts?

### DIFF
--- a/src/backend.js
+++ b/src/backend.js
@@ -26,8 +26,8 @@ var _ = Mavo.Backend = $.Class({
 		}
 	},
 
-	async get (url = new URL(this.url)) {
-		if (url.protocol != "data:") {
+	async get (url = new URL(this.url), {useCache = false} = {}) {
+		if (!useCache && url.protocol != "data:") {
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
It might be just me, but I'm personally leaning towards your suggestion to move `useCache` to the set of backend static properties. It feels like a backend-specific parameter.

LeaVerou:
Agreed.

PR #683, opened 2021-01-25
Spreet2472/refactor

sPreet247:
I have refactored the code base, changed the keyword var to let.
Also added a rule to the eslintrc file to discourage the use of var keyword.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,36 +1,41 @@
 {
-	"parser": "babel-eslint",
-	"parserOptions": {
-		"ecmaVersion": 6,
-		"sourceType": "script",
-		"ecmaFeatures": {
-			"impliedStrict": true
-		}
-	},
-	"env": {
-		"browser": true,
-		"node": true
-	},
-	"rules": {
-		"semi": 1,
-		"no-dupe-args": 1,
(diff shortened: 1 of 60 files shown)


LeaVerou (changes requested):
Hi there,
Thank you for your contribution!
It is rather difficult to review your PR (even with `?w=1`) because of the various other changes, in addition to `var` --> `let`. Please revert the space-based indentation back to smart tabs (tabs for indentation, spaces for alignment). There are also a few more formatting changes I'm unsure about, and it would be good to either remove them, or separate them into a different PR.
For example, I see you have wrapped a lot of lines into multiple lines. We do not slavishly follow a certain line length; we wrap lines of code where it makes the code more readable. Sometimes, a long line can be more readable than a wrapped line, and we leave that up to individual judgement. Note to self: This might be something to add to CONTRIBUTING.md, if it's not there already. (edit: done)

sPreet247:
Hi Lea
I have made the changes you requested.

LeaVerou:
Closing this as it's now very much out of date, and we also don't want to do this migration to `let` in such a bulk way, but gradually.

Sorry about that, and thank you for contributing!

PR #686, opened 2021-02-01
begin moving to let instead of var

### LINKED ISSUE #678: [Chore] Use let instead of var throughout the codebase

MoralCode:
This pull request (theoretically) fixes #678. I pretty much just did a find and replace across the whole codebase and then only committed the changes to files that weren't in `/dist` (since it seemed like those probably shouldn't be modified.

from a quick search, it seems like the difference between let and var is the scope, and this move would change variable scope from their current function to their current set of `{}`. source: https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var

I'll work on running the tests to see if everything still works

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -163,7 +163,9 @@ var _ = Mavo.Actions = {
 				// If to is a number and not a collection item, it's an index
 				[index, to] = [to];
 			}
-
+			//TODO: let->var issue #678
+			//Here is one place where variable hoisting seems to be depended upon with the variable toNode
+			
 			var toNode = _.getNode(to);
 			var fromNodes = Mavo.toArray(from).map(_.getNode).filter(n => n?.closestCollection);
 			var collection = (toNode || fromNodes[0]).closestCollection;
--- a/src/collection.js
+++ b/src/collection.js
@@ -445,7 +445,8 @@ var _ = Mavo.Collection = class Collection extends Mavo.Node {
 				i--;
 			}
 		}
-
+		//TODO: let->var issue #678
+		//it seems like in the for loop above and the if statement below, that the i variable is being re-used from the for loop. this could cause issues in changing it to a let variable
 		if (data.length > i) {
 			// There are still remaining items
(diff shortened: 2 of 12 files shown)


LeaVerou (changes requested):
Hi there, thank you for contributing.
While that is indeed the main difference, there are other subtle differences between `let` and `var` (e.g. `let` variables are not hoisted), so I have to admit that using find-and-replace for this makes me a bit uncomfortable. I would prefer to gradually convert these over time so that the diffs are more manageable and can be reviewed thoroughly. After all there is no pressing need to convert everything at once. One thing to be very careful about is that while in most cases `var` variables are used as block scope anyway and can just be changed, in other cases, they may be defined in certain blocks (e.g. conditionals) and used outside of these blocks. I see you've already taken care of some of that, but thought I'd mention it anyway.

Furthermore:
- We do not modify files in `lib/`, those are dependencies.
- The PR includes some unrelated changes, would I would prefer as a separate PR. 

Btw, we are about to release a new minor version, so I would prefer to avoid introducing such a large commit right before release.

Would you consider breaking this down into multiple (way) smaller PRs?

Thanks again!

MoralCode:
Your comment made me realize that I did not fully think this through beforehand. Sorry for wasting your time.

In an attempt to be as useful as possible with the changes already made:
- i've removed nearly all of the find-and-replace and unrelated changes from this branch
- in the places where I brought a variable "up" to the next scope so it could be used after a conditional as you describe the find-and-replace change was squashed together with the change adjusting the scope.
- I undid the two more complex changes and instead left comments (starting with `//TODO: let->var issue #678` so they are find-able) to document some of the places where there are more complex issues, such as places where `var` scoping behavior is relied upon.

I hope this smaller PR of changes can act as a starting point for this migration, however I understand not wanting to merge in anything too risky right before a release.

Hope this helps and apologies again for wasting your time

LeaVerou:
Closing this as it's now very much out of date, and as explained, we want to do this migration slowly.

PR #694, opened 2021-02-03
[plugins] Let a plugin author pointed out where the plugin hosted

DmitrySharabin:


src/plugins.js
```diff
@@ -34,7 +34,11 @@ var _ = Mavo.Plugins = {
 				// Load plugin
 				var filename = `mavo-${plugin.id}.js`;
 
-				if (plugin.repo) {
+				if (plugin.url) {
+					// The plugin author pointed out where the plugin hosted
```
LeaVerou, line 38: Grammar: "where the plugin is hosted"

src/plugins.js
```diff
@@ -34,7 +34,11 @@ var _ = Mavo.Plugins = {
...
+					var url = plugin.url;
```
LeaVerou, line 39: Let's try and gradually convert `var` to `let` when we touch a piece of code. Especially things like this, which are taking advantage of the fact that `var` is function scope, and will need manual conversion anyway.

LeaVerou (changes requested):
Thanks Dmitry, LGTM, see my two minor comments.

src/plugins.js
DmitrySharabin: Thank you! Fixed.

src/plugins.js
DmitrySharabin: Agreed.

src/plugins.js
```diff
@@ -32,15 +32,20 @@ var _ = Mavo.Plugins = {
...
 
 				// Load plugin
-				var filename = `mavo-${plugin.id}.js`;
+				let filename = `mavo-${plugin.id}.js`;
+				let url;
 
-				if (plugin.repo) {
+				if (plugin.url) {
+					// The plugin author pointed out where the plugin is hosted
+					url = plugin.url;
```
DmitrySharabin, line 40: If we come to the conclusion that `plugin.url` is a base URL for all the plugin files and is not include `filename`, I'll replace this line with:
```js
url = `${plugin.url}/${filename}`
```

LeaVerou: Was there any further discussion on that?

DmitrySharabin: Yep, it's here: https://github.com/mavoweb/plugins/pull/85#discussion_r569444932

DmitrySharabin:
We'll be back to this question later. For now, I'm going to close it.

PR #708, opened 2021-03-20
[Elements] Leftover after migrating to mv-editor*

DmitrySharabin:
Fixes the behavior of the `meter` and the `progress` elements.


LeaVerou:
Whoops, nice catch, thanks!

PR #712, opened 2021-03-30
[Github] Fix issues with forks

DmitrySharabin:
@LeaVerou 

Hi Lea! Will you please take a look at this PR. Here I tried to iron out some bugs when working with the _Switch to my fork_ dialog: as you might have noticed it didn't even appear when an app had both source and storage.

I also sent the corresponding PR to the plugins repo to fix an issue with the broken preview URL—that bug is related to this one. I sent yesterday's PR to the docs via the updated versions of the Github backend and the Github-pr plugin. And you hadn't got any problems with reviewing it, had you? So I hope I could fix the issue. 🤞🏻 If not, I'll continue investigating.

The diff looks a bit strange because I fixed the problem with the indentation in the `switchToMyForkDialog` method as well.

### DIFF
--- a/src/backend.github.js
+++ b/src/backend.github.js
@@ -217,7 +217,12 @@ let _ = Mavo.Backend.register($.Class({
 
 							this.repoInfo = repoInfo;
 
-							if (!this.mavo.source) { // if url doesn't have source, check for forks
+							if (this.mavo.storage) { // Check for forks
+								if (repoInfo.fork) { // Current repo is a fork
+									this.forkInfo = repoInfo.parent;
+									return repoInfo;
+								}
+
 								if (!this.canPush()) { // Check if current user has a fork of this repo, and display dialog to switch
 									if (this.user.info.public_repos < repoInfo.forks) { // graphql search of current user's forks
 										let query = `query {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hi Dmitry, thank you so much for working on this. Could you tell me a little bit about how this code fixes the issue? Is it just about moving code from the plugin to Mavo and vice versa?

DmitrySharabin:
I tried to provide some explanations [here](https://github.com/mavoweb/plugins/pull/88#issuecomment-812072724). Will you please take a look at them first. They concern lines 340 through 357 and are almost identical to the corresponding lines in the `github-pr` plugin.

The rest comments I'll add to the corresponding lines in the diff if you don't mind.

src/backend.github.js
```diff
@@ -217,7 +217,12 @@ let _ = Mavo.Backend.register($.Class({
 
 							this.repoInfo = repoInfo;
 
-							if (!this.mavo.source) { // if url doesn't have source, check for forks
+							if (this.mavo.storage) { // Check for forks
```
DmitrySharabin, line 220: Previously, because of this check, the `Switch to my fork` dialog appeared only when an app hadn't got `source`. And I find it not correct. So I made this change.

src/backend.github.js
```diff
@@ -217,7 +217,12 @@ let _ = Mavo.Backend.register($.Class({
...
+								if (repoInfo.fork) { // Current repo is a fork
```
DmitrySharabin, line 221: I removed this check from the code by mistake when I was moving the code concerning PRs to the plugin. I returned it since it helps to avoid some extra unneeded checks.

DmitrySharabin:
Now, I understand the logic behind the present behavior a bit clearer. So, there is no need for these changes anymore.

Issue #758, opened 2021-08-15
Using an invalid name in mv-bar results in an exception and stack trace


forsyth:
By accident I wrote mv-bar="nologin" instead of mv-bar="no-login" and because I didn't notice the typo, was at first bemused that the page didn't work at all, just defaults, as if my data had vanished. Removing the option restored things so, checking the console, I found the immediate cause:

mavo-nodeps.js:2459 Uncaught (in promise) TypeError: Cannot read property 'create' of undefined
    at mavo-nodeps.js:2459
    at Array.forEach (<anonymous>)
    at n.constructor (mavo-nodeps.js:2451)
    at new n (deps.js:1)
    at new Mavo (mavo-nodeps.js:144)
    at mavo-nodeps.js:827
    at Array.map (<anonymous>)
    at Function.init (mavo-nodeps.js:827)
    at mavo-nodeps.js:1177
(anonymous) @ mavo-nodeps.js:2459
constructor @ mavo-nodeps.js:2451
n @ deps.js:1
Mavo @ mavo-nodeps.js:144
(anonymous) @ mavo-nodeps.js:827
init @ mavo-nodeps.js:827
(anonymous) @ mavo-nodeps.js:1177
async function (async)
(anonymous) @ mavo-nodeps.js:1023
(anonymous) @ mavo-nodeps.js:1180

in 		this.controls.forEach(id => {
			let o = _.controls[id];
                        ...
			if (o.create) {
				this[id] = o.create.call(this.mavo, this[id]);

... _.controls[id] is not set for "nologin" (or nonsense, such as "frodo"). Since there are other references to _.controls[id] it's probably best for getControls to complain and discard useless ones if it can.

LeaVerou:
Confirmed. Testcase: https://codepen.io/leaverou/pen/vYmwYPJ?editors=1100

LeaVerou:
Thank you for reporting this! It should be pretty easy to fix.

LeaVerou:
> ... _.controls[id] is not set for "nologin" (or nonsense, such as "frodo"). Since there are other references to _.controls[id] it's probably best for getControls to complain and discard useless ones if it can.

Just noticed this. Am I reading this correctly that you would be interested in submitting a pull request? 😄

forsyth:
Unfortunately I've only just started using Mavo let alone studied the
internals, and at the moment I'd probably just spoil the existing
cleanliness of the code!

Nisbij:
[...]

PR #760, opened 2021-08-24
resolving #758

Nisbij:
Edited code to resolve issue #768,i.e. not prompting any warning in case of wrong value for the mv-bar attribute


Nisbij:
Can anyone check this out?

PR #775, opened 2021-10-13
Fix build process so that maps point to the correct files

DmitrySharabin:
Let's try to fix the build process so that source maps point to source files, not generated ones.

To make this work I had to remove `sourceMappingURL` from these files:
- `lib/jsep.iife.min.js`
- `lib/jsep.min.js`
- `lib/stretchy.min.js`

It looks like we will have to repeat the same steps every time we update dependencies.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,9 @@ var notify = require("gulp-notify");
 var merge = require("merge2");
 var injectVersion = require("gulp-inject-version");
 var csso = require("gulp-csso");
+var fs = require("fs");
+var sorcery = require("sorcery");
+var replace = require("gulp-replace");
 
 sass.compiler = require("sass");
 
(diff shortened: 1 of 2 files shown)


LeaVerou:
We cannot be modifying build artifacts. This is going to be a maintenance nightmare. If this is absolutely necessary, our build process should do it automatically before using the file.

Thanks for working on this btw! It has been bugging me for YEARS!

DmitrySharabin:
I think I managed to fix it. 🤞🏻

@LeaVerou could you please give it a try and let me know if there is something else I need to fix?

DmitrySharabin:
> It seems a little roundabout to generate incorrect maps and then fix them, and I have a hunch that there is a simpler solution where we just generate correct maps in the first place.

Right you are! I'll try to find a better solution a bit later. Thanks for merging!

PR #783, opened 2021-10-19
[Github] Minor fix so that we cover all kinds of raw API calls

DmitrySharabin:
It will also remove stray `console.log()`.


LeaVerou:
Thanks!

PR #796, opened 2021-11-25
Tweak the minifier so that it works correctly with ES6 classes

DmitrySharabin:
There were two issues with the minifier:
– it was removing class names and
– was merging variable declarations so that sometimes the variables might be left undefined (merging might clash with other code transformations)


DmitrySharabin:
Since Lea encourages me to commit this kind of change directly, I'm going to take this opportunity. 😜

PR #799, opened 2021-12-01
[mavo] Some fixes

DmitrySharabin:
I looked carefully through `mavo.js` and found a couple of things to improve.

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -626,7 +626,7 @@ let _ = self.Mavo = $.Class(class Mavo {
 			return Promise.reject();
 		}
 
-		this.inProgress = this._("uploading");
+		this.inProgress = "Uploading";
 
 		return this.uploadBackend.upload(file, path)
 			.then(url => {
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
I would also suggest replacing this line:
https://github.com/mavoweb/mavo/blob/0ee785f706223d627debb040ed5741a2ba094369/src/mavo.js#L629
with this one:
```js

this.inProgress = "Uploading";
```
for the sake of consistency and to let authors use the `[mv-progress= "Uploading"]` in their stylesheets if they need it.

If we are OK with that, I’m ready to push the corresponding commit.

LeaVerou:
> I would also suggest replacing this line:
> 
> https://github.com/mavoweb/mavo/blob/0ee785f706223d627debb040ed5741a2ba094369/src/mavo.js#L629
> 
> 
> with this one:
> ```js
> this.inProgress = "Uploading";
> ```
> 
> for the sake of consistency and to let authors use the `[mv-progress= "Uploading"]` selector in their stylesheets if they need it.
> 
> If we are OK with that, I’m ready to push the corresponding commit.

Yes, that makes sense.

LeaVerou (approved):

PR #801, opened 2021-12-02
[actions] Some enhancements 

DmitrySharabin:
In this PR I propose a fix for the bug we spotted yesterday and replaced `var` with `let`. It shouldn't break anything (if I haven't screwed anything up 🤞🏻). Actions tests still pass.

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -170,12 +170,13 @@ var _ = Mavo.Actions = {
 				return;
 			}
 
+			var toNode = _.getNode(to);
+
 			if ($.type(to) == "number" && !(toNode?.collection)) {
 				// If to is a number and not a collection item, it's an index
 				[index, to] = [to];
 			}
 
-			var toNode = _.getNode(to);
 			var fromNodes = Mavo.toArray(from).map(_.getNode).filter(n => n?.closestCollection);
 			var collection = (toNode || fromNodes[0]).closestCollection;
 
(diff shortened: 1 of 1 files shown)


src/actions.js
```diff
@@ -170,14 +170,15 @@ var _ = Mavo.Actions = {
 				return;
 			}
 
+			let toNode = _.getNode(to);
+
 			if ($.type(to) == "number" && !(toNode?.collection)) {
 				// If to is a number and not a collection item, it's an index
 				[index, to] = [to];
```
LeaVerou, line 177: Note that this can change `to`, so now that `toNode` is set before this, they may end up getting out of sync. However, this essentially deletes `to`, so I guess all you need to do to `toNode` is setting it to `undefined` (but I haven't tried it, obvs)

DmitrySharabin: I need to look at it more carefully then. Thanks.

DmitrySharabin: Thanks for the suggestion, Lea. You were right. I committed the corresponding change.

Additionally to tests, I also checked our demos, docs, and some of my apps. I haven't found any regressions.

DmitrySharabin:
@LeaVerou

Hey Lea, 👋🏻

Is there a chance you could give this PR another look when you have time? I made the change you proposed.

LeaVerou (approved):

LeaVerou:
LGTM!

Issue #603, opened 2020-06-12
[plugins] jsDelivr doesn't deliver the latest plugin version


DmitrySharabin:
I released a new version of the Firebase plugin. But when I include it via the `mv-plugins` attribute, jsDelivr delivers me the old version. Do we have to wait a bit for changes to propagate? What if we use the `https://cdn.jsdelivr.net/gh/${plugin.repo}@latest/${filename}` URL to include plugins instead? For not to wait.

DmitrySharabin:
Yes, we have to wait for the changes to propagate even if we use the `@latest` suffix (so there is no need to use it). I believe we should warn plugins authors somehow about this jsDelivr “feature” and not make them panic as I was. 😂

LeaVerou:
How long is the wait?

DmitrySharabin:
It feels like 12–24 hours.

LeaVerou:
Oh wow, that's a lot.

LeaVerou:
It looks like jsdelivr might only work with releases, e.g. this link produces a 404: https://cdn.jsdelivr.net/gh/valterkraemer/mavo-firebase/mavo-firebase.js

and the file is not in the directory listing: https://cdn.jsdelivr.net/gh/valterkraemer/mavo-firebase/

It looks like we may need to find an alternative if there's no jsdelivr syntax that will give us the latest commit. @DmitrySharabin any ideas?

DmitrySharabin:
To my mind, we might have two options here.

Plugin authors are skilled enough so they can enable **GitHub Pages**. If they haven’t done it before, it won't be a big problem for us to give them a link where they can find full instructions on how to do that. So we can deliver plugins from GitHub directly.

Since plugin authors can use custom domains for GitHub Pages or host their plugins somewhere else, we can provide an optional parameter (property) for plugins, e.g., `pluginURL`. If it's empty, we use the default URL—`https://author.github.io/plugin-repo/mavo-plugin-id.js`. Otherwise, we use the URL provided by a plugin’s author.

What do you think?

LeaVerou:
Yeah, I was wondering about requiring Github Pages. But this doesn't solve anything about existing plugins, many of which are unmaintained so we can't chase down their authors and force them to enable Github Pages. 
Btw I just pushed a fix that fixes the URL to get the latest commit on master, I wonder if this also addresses the long wait?

LeaVerou:
[...]

PR #803, opened 2021-12-13
Allow for specific versions of plugins to be used

DmitrySharabin:
We discussed this feature once here: #603. And I decided to give it a shot.

For now, it works only for plugins hosted in a separate repo. I'm not sure whether it's possible to release our plugins separately (since all of them are in the same repo).

I also bumped into an issue: it looks like we download all versions of a plugin even though we expect that the first specified version wins (in the order the plugins are specified in `mv-plugins`). Why do we expect this? Because of this line:
```js
if (_.loaded[plugin.id]) {
	return Promise.resolve();
}
```
However, it looks like we add plugins to the `loaded` object too late (inside `Mavo.Plugins.registered()`).

Are there any thoughts on how to fix it?

### DIFF
--- a/src/plugins.js
+++ b/src/plugins.js
@@ -2,7 +2,7 @@
 
 Mavo.attributes.push("mv-plugins");
 
-var _ = Mavo.Plugins = {
+let _ = Mavo.Plugins = {
 	loaded: {},
 
 	async load () {
@@ -32,15 +32,16 @@ var _ = Mavo.Plugins = {
 				}
 
 				// Load plugin
-				var filename = `mavo-${plugin.id}.js`;
+				let filename = `mavo-${plugin.id}.js`;
+				let url;
 
 				if (plugin.repo) {
 					// Plugin hosted in a separate repo
-					var url = `https://cdn.jsdelivr.net/gh/${plugin.repo}@master/${filename}`;
+					url = `https://cdn.jsdelivr.net/gh/${plugin.repo}@master/${filename}`;
 				}
 				else {
 					// Plugin hosted in the mavo-plugins repo
-					var url = `${_.url}/${plugin.id}/${filename}`;
+					url = `${_.url}/${plugin.id}/${filename}`;
 				}
 
 				return $.include(_.loaded[plugin.id], url);
@@ -66,15 +67,15 @@ var _ = Mavo.Plugins = {
 			}
 		}
 
-		var ready = [];
+		let ready = [];
 
 		if (o.ready) {
 			ready.push(o.ready);
 		}
 
 		if (o.dependencies) {
-			var base = document.currentScript? document.currentScript.src : location;
-			var dependencies = o.dependencies.map(url => Mavo.load(url, base));
+			let base = document.currentScript? document.currentScript.src : location;
+			let dependencies = o.dependencies.map(url => Mavo.load(url, base));
 			ready.push(...dependencies);
 		}
 
--- a/src/plugins.js
+++ b/src/plugins.js
@@ -8,10 +8,19 @@ let _ = Mavo.Plugins = {
 	async load () {
 		_.plugins = new Set();
 
+		let versions = {};
 		$$("[mv-plugins]").forEach(element => {
 			element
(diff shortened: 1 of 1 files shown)


LeaVerou:
> We discussed this feature once here: #603. And I decided to give it a shot.
> 
> For now, it works only for plugins hosted in a separate repo. I'm not sure whether it's possible to release our plugins separately (since all of them are in the same repo).

What a great feature! Have you tested it works well even for plugins that are in a separate repo and have *never* had a release?

For plugins in our own repo, we should move towards a monorepo structure. I wonder if that allows separate versions? Otherwise I guess we'd need to create a new release for any substantial change to any plugin.
Regardless, the feature should still work for plugins in our repo, and just refer to versions of our repo.

> I also bumped into an issue: it looks like we download all versions of a plugin even though we expect that the first specified version wins (in the order the plugins are specified in `mv-plugins`). Why do we expect this? Because of this line:
> 
> ```js
> if (_.loaded[plugin.id]) {
> 	return Promise.resolve();
> }
> ```
> 
> However, it looks like we add plugins to the `loaded` object too late (inside `Mavo.Plugins.registered()`).
> 
> Are there any thoughts on how to fix it?

I don't understand the question, possibly because my brain is fried right now. Where is that code and why does it download multiple versions?

DmitrySharabin:
>  Have you tested it works well even for plugins that are in a separate repo and have never had a release?

Yup. I took into account the fact that if there are no releases in a repo, `@latest` fallbacks to the default branch (either `master` or `main`). You can follow these links to see it in action:
- https://cdn.jsdelivr.net/gh/dmitrysharabin/mavo-locale-ru@latest/mavo-locale-ru.js
- https://cdn.jsdelivr.net/gh/dmitrysharabin/mavo-locale-ru/mavo-locale-ru.js

Those are files from my repo without releases (the `locale-ru` plugin): https://github.com/DmitrySharabin/mavo-locale-ru.

> I don't understand the question, possibly because my brain is fried right now. 

It's my poor wording. I'll prepare a better explanation. :)

DmitrySharabin:
> I don't understand the question, possibly because my brain is fried right now.

Unfortunately, that was not only my poor wording but also not very thoughtful logic. Luckily, I could mend my ways and get not only the shorter but more proof code (I hope so 🤞🏻). I documented the tests I performed. Did I miss anything?

@LeaVerou Lea, could you please give it another look when you have time?

## Tests

### Default
```html
<div mv-app mv-plugins="locale-ru chart tinymce">
```

**Result**: loaded the latest versions.

### Non-existent
```html
<div mv-app mv-plugins="non-existent">
```

**Result**: no files loaded.

### Repo with no releases
```html
<div mv-app mv-plugins="locale-ru">
```

**Result**: loaded the latest version.

### No such release
```html
<div mv-app mv-plugins="chart@10.0.3 firebase-firestore@foo">
```

**Result**: loaded the latest versions.

### The last one wins
```html
<div mv-app mv-plugins="firebase-firestore@0.3.9 firebase-firestore firebase-firestore@0.3.8">
```
**Result**: loaded version `v0.3.8`.

### The last one wins + no such release
```html
<div mv-app mv-plugins="firebase-firestore@0.3.9 firebase-firestore firebase-firestore@10.0.3">
```

**Result**: loaded the latest version.

### Release with `v` before the number and without (either work)
```html
<div mv-app mv-plugins="firebase-firestore@v0.3.9">
```

**Result**: loaded version `v0.3.9`.

### Versions of our plugins
```html
<div mv-app mv-plugins="tinymce@v1 dropbox@42">
```

**Result**: loaded the latest versions.

LeaVerou:
Great work Dmitry! It would be nice to be able to do partial versions (e.g. `foo@10`) but that can wait for a future PR.

LeaVerou (approved):
Fantastic work

DmitrySharabin:
> It would be nice to be able to do partial versions (e.g. foo@10) but that can wait for a future PR.

Luckily, we already have it, thanks to the way jsdelivr works. 🥳 Added the corresponding tests to the comment above.

> Fantastic work

Thank you so much!!

PR #805, opened 2021-12-18
Clicking on editable property inside <a> mustn't trigger the link

DmitrySharabin:
Let's try to fix this issue one more time.

This fix doesn't cover this case. But it looks so weird. I can't even imagine when we need it. But maybe it's just me. 😅
```html
<a href="foo.html"><span property="yolo" mv-mode="edit">yolo</span></a>
```

I tested all the known cases and regressions (including the fresh one you built) and checked the test suite. It looks like everything works as expected. 🤞🏻

### DIFF
--- a/src/primitive.js
+++ b/src/primitive.js
@@ -449,14 +449,14 @@ var _ = Mavo.Primitive = class Primitive extends Mavo.Node {
 	}
 
 	edit (o = {}) {
-		let wasEditing = this.editing;
+		let wasEditing = this.editing && !this.initEdit;
 
 		if (super.edit(o) === false) {
 			// Invalid edit
 			return false;
 		}
 
-		if (!o.force && wasEditing && !this.initEdit) {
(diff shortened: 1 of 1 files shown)


LeaVerou (commented):
I have some reservations that this will actually address all the problematic cases, though I’m tempted to merge it because I'm sure it's better than what we have right now.

Have you considered cases like these:

```html
<a href>
	<span property>
</a>
```

```html
<span property>
	text
	<a href></a>
</span>
```

etc?

What about other default actions besides `<a>`, like `<summary>` toggling or form submission?

We should have tests explicitly for this. With `data-click` they could even be automated.

DmitrySharabin:
> I have some reservations that this will actually address all the problematic cases

Me too. I'll begin with tests then (you mentioned cases I haven't even thought about 😔). That will be a much better approach than the one I use now. I'm afraid you shouldn't approve and merge this PR yet. I'll work on it more.

Issue #793, opened 2021-10-28
Consistent function naming: let's go with underscores


LeaVerou:
Currently we have very few MS functions with 2 word names (thanks @DmitrySharabin for hunting them down):

* `groupby()`
* `tolast()`
* `fromlast()`

Given that Mavo is generally case insensitive, I don't think concatenating names makes for very readable code. Instead, we should use underscores to separate words, which is guaranteed to be readable with any casing.

A solid plan for this would be:

* Any new multi-word functions use underscores (e.g. [`readable_datetime()`](https://github.com/mavoweb/mavo/commit/a8c5e827c969349bc7cb03ec7c0e9e86e2fdefe2))
* We rename the three functions above to `group_by()`, `to_last()`, `from_last()`

We add the current names as aliases, which call `Mavo.warn()` to print out something like: 

```
groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.
```

and then call the new function name.

We need to also make sure we're not calling the old names in any of our own code or plugins.

amm98d:
> > Is this up-for-grabs? Would love to contribute on this :)
> 
> Sure, thanks! One caveat is that while most of these are defined in `src/functions.js`, the `groupby()` function is defined in `src/mavoscript.js` because it's also an operator.

Alright, will send PR soon !

jvitor211:
Hi I would like to contribute, are you still open for help?

LeaVerou:
> Hi I would like to contribute, are you still open for help?

Sure!
How did you find this project, if I may ask?

jvitor211:
a few days ago I was on the csail MIT website and found this research group
and commented with the professor, Mr. David Karger.
And I talked about the project asking if I could be part of it in any way
and contribute

https://prnt.sc/1y67p4j
https://prnt.sc/1y67gjx

Em ter., 2 de nov. de 2021 às 09:58, Lea Verou ***@***.***>
escreveu:

NEERAJthinker:
I would like to contribute on this. This is going to be my first contribution, therefore can you please tell me where to start.

norbitrial:
> Currently we have very few MS functions with 2 word names (thanks @DmitrySharabin for hunting them down):
> 
> * `groupby()`
> * `tolast()`
> * `fromlast()`
>
> Given that Mavo is generally case insensitive, I don't think concatenating names makes for very readable code. Instead, we should use underscores to separate words, which is guaranteed to be readable with any casing.

[...]

PR #821, opened 2022-01-23
feat: consistent function naming with underscores

norbitrial:
This PR is a solution for #793 issue.

Function calls are tested after renaming and working exact the same way as before. There are extra warnings on the console for old function calls for the following cases:
- `groupby()` function call,
- `fromlast()` and `tofirst()` calls.

Let me know if you have any suggestions. Thank you! 🙇 👍

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -492,7 +492,6 @@ var _ = Mavo.Script = {
 			precedence: 2
 		},
 		"groupby": {
-			symbol: "by",
 			code: (array, key) => {
 				Mavo.warn("groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.");
 				return Mavo.Functions["group_by"](array, key);
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -453,14 +453,6 @@ var _ = Mavo.Script = {
 			},
 			precedence: 3
 		},
-		"groupby": {
-			symbol: "by",
-			code: (array, key) => {
-				Mavo.warn("groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.");
-				Mavo.group_by(array, key);
-			},
-			precedence: 2
-		},
 		"group_by": {
 			symbol: "by",
 			code: (array, key) => {
@@ -496,6 +488,14 @@ var _ = Mavo.Script = {
 				});
 
 				return Mavo.Data.proxify(ret);
+			},
+			precedence: 2
+		},
+		"groupby": {
+			symbol: "by",
+			code: (array, key) => {
+				Mavo.warn("groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.");
+				return Mavo.Functions['group_by'](array, key);
 			},
 			precedence: 2
 		},
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -495,7 +495,7 @@ var _ = Mavo.Script = {
 			symbol: "by",
 			code: (array, key) => {
 				Mavo.warn("groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.");
-				return Mavo.Functions['group_by'](array, key);
+				return Mavo.Functions["group_by"](array, key);
 			},
 			precedence: 2
 		},
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -492,7 +492,7 @@ var _ = Mavo.Script = {
 			precedence: 2
 		},
 		"groupby": {
-			code: Mavo.deprecatedFunction("group_by", "groupby"),
+			code: $u.deprecatedFunction("group_by", "groupby"),
 			precedence: 2
 		},
 		"as": {
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -456,6 +456,14 @@ var _ = Mavo.Script = {
 		"groupby": {
 			symbol: "by",
 			code: (array, key) => {
+				Mavo.warn("groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.");
+				Mavo.group_by(array, key);
+			},
+			precedence: 2
+		},
+		"group_by": {
+			symbol: "by",
+			code: (array, key) => {
 				array = Mavo.toArray(array);
 				key = Mavo.toArray(key);
 				var property = key[Mavo.as] || key[0]?.[Mavo.toNode]?.property;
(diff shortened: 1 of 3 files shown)


src/mavoscript.js
```diff
@@ -491,6 +491,14 @@ var _ = Mavo.Script = {
 			},
 			precedence: 2
 		},
+		"groupby": {
+			symbol: "by",
```
DmitrySharabin, line 495: I believe we don't need this line anymore since after the PR is merged, `a by b` will be the same as `group_by(a, b)`, and not `group(a, b)`. Do you mind trying to remove it and see what we get?

src/mavoscript.js
```diff
@@ -491,6 +491,14 @@ var _ = Mavo.Script = {
...
+			code: (array, key) => {
+				Mavo.warn("groupby() is deprecated and will be removed in the next version of Mavo. Please use group_by() instead.");
+				return Mavo.Functions['group_by'](array, key);
```
DmitrySharabin, line 498:
```suggestion
				return Mavo.Functions["group_by"](array, key);
```

To be consistent with the rest of the codebase.

DmitrySharabin (changes requested):
That's awesome! Thank you. Actually, you already did a great job; there is only one question left. Please see the comments.

DmitrySharabin (changes requested):
That's awesome! Thank you. Actually, you already did a great job; there is only one question left. Please see the comments.

DmitrySharabin:
By the way, it would be nice if you could also make the corresponding changes to our [testsuite](https://mavo.io/docs/functions) so that it reflects the changes to the API. To be precise, to [this file](https://github.com/mavoweb/test/blob/master/functions.html).

LeaVerou:
> By the way, it would be nice if you could also make the corresponding changes to our [testsuite](https://mavo.io/docs/functions) so that it reflects the changes to the API. To be precise, to [this file](https://github.com/mavoweb/test/blob/master/functions.html?rgh-link-date=2023-03-07T21%3A06%3A32Z). But that's not necessary. I can do it too. 😀

Actually, it's even more important to make the changes in our docs. (which should also include both versions, with e.g. `groupBy` just saying something like "Name of `group_by()` before v0.3.0", with a note about which Mavo version supports what)

DmitrySharabin:
@LeaVerou, it seems to me this PR is also worth merging before `v0.3.0` is released. It will allow us to be consistent with the new functions (e.g., `datetime_readable()`). @norbitrial did a great job. There is a couple of minor issues left I can fix myself.

LeaVerou (changes requested):
LGTM, my only suggestion would be to abstract the deprecation code away into a function instead of repeating it. 

Something like:

```js
function deprecatedFunction(name, oldName, fn = _[name]) {
	return function (...args) {
		Mavo.warn(`${oldName}() is deprecated and will be removed in the next version of Mavo. Please use ${name}() instead.");
		return fn(...args);
	}
}
```

...and to make the docs edits!

In fact, in https://mavo.io/docs/functions/ I think we should have two entries for each of these, with the old one saying "Deprecated alias of xxx()"

DmitrySharabin:
Done.

LeaVerou (changes requested):
Can we not remove extra commas please? 

a) trailing commas are there for a reason, it means that adding things after is less likely to result in a syntax error
b) even if we wanted to remove trailing commas, this PR is not the place and it increases the size of the diff making it harder to review

Also, this abstracted away the deprecation code, but is not using it in all three places where it's needed.
We should expose it somewhere, so that other modules can use it as well (or even plugins).

DmitrySharabin:
> Can we not remove extra commas please?

Sorry about that. I thought they were some leftovers from a code formatter (e.g., Prettier). Fixed it.

> Also, this abstracted away the deprecation code, but is not using it in all three places where it's needed. We should expose it somewhere, so that other modules can use it as well (or even plugins).

I hesitated to put it into `utils.js` and made the wrong choice. I should have asked. From my point of view, `utils.js` might be a good place for the `deprecatedFunction()` function.

LeaVerou:
> > Can we not remove extra commas please?
> 
> Sorry about that. I thought they were some leftovers from a code formatter (e.g., Prettier). Fixed it.
> 
> > Also, this abstracted away the deprecation code, but is not using it in all three places where it's needed. We should expose it somewhere, so that other modules can use it as well (or even plugins).
> 
> I hesitated to put it into `utils.js` and made the wrong choice. I should have asked. From my point of view, `utils.js` might be a good place for the `deprecatedFunction()` function. If not, I'm open to making other changes if needed.

Yeah, I can't think of a better place either, so utils it is. Or `Mavo.Script`. Up to you!

DmitrySharabin:
> Yeah, I can't think of a better place either, so utils it is. Or Mavo.Script. Up to you!

Then, I believe it's ready for your review. Again. 😅

LeaVerou:
Actually I know where it should live! On `Mavo.Functions.util`! Apologies, I forgot about that entirely.

DmitrySharabin:
> Actually I know where it should live! On `Mavo.Functions.util`! Apologies, I forgot about that entirely.

Honestly, I was also trying to “house” `deprecatedFunction()` there (it was one of my first attempts). However, I faced an issue I didn't (and still don't) know how to resolve elegantly.

Suppose we have the following code (where `$u` is an alias for `_.util`):

```js
fromlast: $u.deprecatedFunction("from_last", "fromlast")
```
And since `util` is defined below and we are trying to use something that is not yet defined, we get an error `TypeError: Cannot read properties of undefined (reading 'util')`.

I suspect that we can do something like this (and it works)

```js
fromlast: (...args) => $u.deprecatedFunction("from_last", "fromlast").call(this, ...args)
```
But I'm unsure whether it's the correct way to do this. What am I missing?

LeaVerou:
> > Actually I know where it should live! On `Mavo.Functions.util`! Apologies, I forgot about that entirely.
> 
> Honestly, I was also trying to “house” `deprecatedFunction()` there (it was one of my first attempts). However, I faced an issue I didn't (and still don't) know how to resolve elegantly.
> 
> Suppose we have the following code (where `$u` is an alias for `_.util`):
> 
> ```js
> fromlast: $u.deprecatedFunction("from_last", "fromlast")
> ```
> 
> And since `util` is defined below and we are trying to use something that is not yet defined, we get an error `TypeError: Cannot read properties of undefined (reading 'util')`.
> 
> I suspect that we can do something like this (and it works)
> 
> ```js
> fromlast: (...args) => $u.deprecatedFunction("from_last", "fromlast").call(this, ...args)
> ```
> 
> But I'm unsure whether it's the correct way to do this. What am I missing?

Just move it up?

DmitrySharabin:
> Just move it up?

It worked! 🥳 Thank you.

LeaVerou (approved):

DmitrySharabin:
@norbitrial, thank you so much for your work!

PR #833, opened 2022-04-18
A way for attributes to contain expressions without triggering browser errors or false loading

### LINKED ISSUE #475: Provide a way for attributes to contain expressions without triggering browser errors or false loading
While allowing expressions in attributes is certainly convenient, in certain cases it has annoying side effects. For example, putting expressions in `img/@src` fires a pointless HTTP request, before the expression has a chance to resolve. Also, putting expressions in attributes that accept restricted values (e.g. numbers) results in pointless errors. 
The [SVG Path builder demo](https://mavo.io/demos/svgpath/) is a great example of this:
image

We should support an alternative syntax for these cases. E.g. (using src as an example) something like `mv-attr-src`, or `mv-@src` or even just `@src` (alternative suggestions welcome!)
This has the added benefit of allowing fallbacks by just providing a value on the actual attribute.
This will also make Viviz's use case easier, since it will enable it to create SVGs that can be read by any SVG editor but will *still* have their expressions.
If we leave the original attribute (`mv-attr-src` or whatever) intact, this can also greatly help SSR (cc @betaveros who may find this interesting).

Whoever implements this, one caveat: `` should still be read-only. Depending on how this is implemented, it may or may not require changes in `Mavo.Primitive`. Assuming the attribute is rewritten in `Mavo.DOMExpression` (like how `mv-value` works), then it should work automatically.

DmitrySharabin:
Closes #475.

I believe this feature is a nice to have in the upcoming release. I tested it on several examples, including [SVG Path Builder](https://mavo.io/demos/svgpath/). It looks like it's working. :)

I liked the idea of using the `mv-attr-*` family of attributes for the feature, so I went that way.

Since I'm not 100% sure that I do everything correctly, let's consider this PR as the first step. :)

### DIFF
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -35,7 +35,7 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.originalAttribute = this.attribute;
 			this.attribute = this.attribute.replace("mv-attr-", "");
 
-			if (this.element.namespaceURI === "http://www.w3.org/2000/svg") {
+			if (["http://www.w3.org/2000/svg", "http://www.w3.org/1998/Math/MathML"].includes(this.element.namespaceURI)) {
 				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
 			}
 
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -43,6 +43,10 @@ var _ = Mavo.DOMExpression = $.Class({
 			}
 
 			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
+			let expression = this.element.getAttribute(this.originalAttribute);
+			this.element.removeAttribute(this.originalAttribute);
+			this.parsed = [new Mavo.Expression(expression)];
+			this.expression = expression;
 		}
 
 		if (this.node.nodeType === 3 && this.element === this.node) {
@@ -62,14 +66,7 @@ var _ = Mavo.DOMExpression = $.Class({
 
 		if (typeof this.expression !== "string") { // Still unhandled?
 			if (this.attribute) {
-				let value;
-				if (this.originalAttribute?.startsWith("mv-attr-")) {
-					value = originalGetAttribute.call(this.node, this.originalAttribute);
-					this.node.removeAttribute(this.originalAttribute);
-				}
-				else {
-					value = originalGetAttribute.call(this.node, this.attribute);
-				}
+				let value = originalGetAttribute.call(this.node, this.attribute);
 
 				this.expression = (value || "").trim();
 			}
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -60,13 +60,15 @@ var _ = Mavo.DOMExpression = $.Class({
 		if (typeof this.expression !== "string") { // Still unhandled?
 			if (this.attribute) {
 				// Some web components (e.g. AFrame) hijack getAttribute()
+				const originalGetAttribute = Element.prototype.getAttribute;
+
 				let value;
 				if (this.originalAttribute?.startsWith("mv-attr-")) {
-					value = Element.prototype.getAttribute.call(this.node, this.originalAttribute);
+					value = originalGetAttribute.call(this.node, this.originalAttribute);
 					this.node.removeAttribute(this.originalAttribute);
 				}
 				else {
-					value = Element.prototype.getAttribute.call(this.node, this.attribute);
+					value = originalGetAttribute.call(this.node, this.attribute);
 				}
 
 				this.expression = (value || "").trim();
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -34,7 +34,11 @@ var _ = Mavo.DOMExpression = $.Class({
 		if (this.attribute?.startsWith("mv-attr-")) {
 			this.originalAttribute = this.attribute;
 			this.attribute = this.attribute.replace("mv-attr-", "");
-			this.attribute = Mavo.adjustSVGAttribute(this.attribute);
+
+			if (this.element.namespaceURI === "http://www.w3.org/2000/svg") {
+				this.attribute = Mavo.getProperCasing(this.element.nodeName, this.attribute);
+			}
+
 			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
 		}
 
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -36,7 +36,7 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.attribute = this.attribute.replace("mv-attr-", "");
 
 			if (this.element.namespaceURI === "http://www.w3.org/2000/svg") {
-				this.attribute = Mavo.getProperAttributeCase(this.element.nodeName, this.attribute);
+				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
 			}
 
 			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -1,5 +1,8 @@
 (function($, $$) {
 
+// Some web components (e.g. AFrame) hijack getAttribute()
+const originalGetAttribute = Element.prototype.getAttribute;
+
 var _ = Mavo.DOMExpression = $.Class({
 	async constructor (o = {}) {
 		this.mavo = o.mavo;
@@ -59,9 +62,6 @@ var _ = Mavo.DOMExpression = $.Class({
 
 		if (typeof this.expression !== "string") { // Still unhandled?
 			if (this.attribute) {
-				// Some web components (e.g. AFrame) hijack getAttribute()
-				const originalGetAttribute = Element.prototype.getAttribute;
-
 				let value;
 				if (this.originalAttribute?.startsWith("mv-attr-")) {
 					value = originalGetAttribute.call(this.node, this.originalAttribute);
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -27,22 +27,22 @@ var _ = Mavo.DOMExpression = $.Class({
 
 		Mavo.hooks.run("domexpression-init-start", this);
 
-		if (this.attribute == "mv-value") {
-			this.originalAttribute = "mv-value";
-			this.attribute = Mavo.Primitive.getValueAttribute(this.element);
-			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
-			let expression = this.element.getAttribute("mv-value");
-			this.element.removeAttribute("mv-value");
-			this.parsed = [new Mavo.Expression(expression)];
-			this.expression = expression;
-		}
-
-		if (this.attribute?.startsWith("mv-attr-")) {
+		if (/^mv-(value$|attr-)/.test(this.attribute)) {
+			// Attributes transformed to other attributes
 			this.originalAttribute = this.attribute;
-			this.attribute = this.attribute.replace("mv-attr-", "");
 
-			if ([SVG_NAMESPACE_URI, MATHML_NAMESPACE_URI].includes(this.element.namespaceURI)) {
-				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
+			if (this.attribute == "mv-value") {
+				this.attribute = Mavo.Primitive.getValueAttribute(this.element);
+			}
+			else {
+				this.attribute = this.attribute.replace("mv-attr-", "");
+
+				// Get proper attribute case if in a case sensitive environment
+				// FIXME do we also need this for XHTML?
+				// FIXME what about namespaced attributes? (e.g. xlink:href)
+				if ([SVG_NAMESPACE_URI, MATHML_NAMESPACE_URI].includes(this.element.namespaceURI)) {
+					this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
+				}
 			}
 
 			this.fallback ??= Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -36,7 +36,7 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.attribute = this.attribute.replace("mv-attr-", "");
 
 			if (this.element.namespaceURI === "http://www.w3.org/2000/svg") {
-				this.attribute = Mavo.getProperCasing(this.element.nodeName, this.attribute);
+				this.attribute = Mavo.getProperAttributeCase(this.element.nodeName, this.attribute);
 			}
 
 			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -3,6 +3,9 @@
 // Some web components (e.g. AFrame) hijack getAttribute()
 const originalGetAttribute = Element.prototype.getAttribute;
 
+const SVG_NAMESPACE_URI = "http://www.w3.org/2000/svg";
+const MATHML_NAMESPACE_URI = "http://www.w3.org/1998/Math/MathML";
+
 var _ = Mavo.DOMExpression = $.Class({
 	async constructor (o = {}) {
 		this.mavo = o.mavo;
@@ -38,7 +41,7 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.originalAttribute = this.attribute;
 			this.attribute = this.attribute.replace("mv-attr-", "");
 
-			if (["http://www.w3.org/2000/svg", "http://www.w3.org/1998/Math/MathML"].includes(this.element.namespaceURI)) {
+			if ([SVG_NAMESPACE_URI, MATHML_NAMESPACE_URI].includes(this.element.namespaceURI)) {
 				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
 			}
 
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -120,17 +120,7 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.item.expressions.delete(this);
 		}
 
-		let expressions = [...this.item.expressions].filter(e => e.attribute === this.attribute && e.active !== false);
-		if (expressions.some(e => e.originalAttribute?.startsWith("mv-attr-"))) {
-			for (let expression of expressions.filter(e => !e.originalAttribute?.startsWith("mv-attr-"))) {
-				// We already have expressions associated with this attribute; discard them since mv-attr-* has priority
-				expression.active = false;
-			}
-		}
-
-		if (this.active !== false) {
-			this.mavo.expressions.register(this);
-		}
+		this.mavo.expressions.register(this);
 
 		Mavo.hooks.run("domexpression-init-treebuilt", this);
 	},
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -45,7 +45,7 @@ var _ = Mavo.DOMExpression = $.Class({
 				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
 			}
 
-			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
+			this.fallback ??= Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
 			let expression = this.element.getAttribute(this.originalAttribute);
 			this.element.removeAttribute(this.originalAttribute);
 			this.parsed = [new Mavo.Expression(expression)];
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -34,6 +34,7 @@ var _ = Mavo.DOMExpression = $.Class({
 		if (this.attribute?.startsWith("mv-attr-")) {
 			this.originalAttribute = this.attribute;
 			this.attribute = this.attribute.replace("mv-attr-", "");
+			this.attribute = Mavo.adjustSVGAttribute(this.attribute);
 			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
 		}
 
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -120,7 +120,17 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.item.expressions.delete(this);
 		}
 
-		this.mavo.expressions.register(this);
+		let expressions = [...this.item.expressions].filter(e => e.attribute === this.attribute && e.active !== false);
+		if (expressions.some(e => e.originalAttribute?.startsWith("mv-attr-"))) {
+			for (let expression of expressions.filter(e => !e.originalAttribute?.startsWith("mv-attr-"))) {
+				// We already have expressions associated with this attribute; discard them since mv-attr-* has priority
+				expression.active = false;
+			}
+		}
+
+		if (this.active !== false) {
+			this.mavo.expressions.register(this);
+		}
 
 		Mavo.hooks.run("domexpression-init-treebuilt", this);
 	},
--- a/src/domexpression.js
+++ b/src/domexpression.js
@@ -21,19 +21,20 @@ var _ = Mavo.DOMExpression = $.Class({
 
 		Mavo.hooks.run("domexpression-init-start", this);
 
-		if (this.attribute === "mv-value" || this.attribute?.startsWith("mv-attr-")) {
-			let replaceableAttribute;
-			if (this.attribute.startsWith("mv-attr-")) {
-				replaceableAttribute = this.attribute.replace("mv-attr-", "");
-			}
-
-			this.originalAttribute = this.attribute;
-			this.attribute = replaceableAttribute ?? Mavo.Primitive.getValueAttribute(this.element);
+		if (this.attribute == "mv-value") {
+			this.originalAttribute = "mv-value";
+			this.attribute = Mavo.Primitive.getValueAttribute(this.element);
 			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
-			let expression = this.element.getAttribute(this.originalAttribute);
-			this.element.removeAttribute(this.originalAttribute);
+			let expression = this.element.getAttribute("mv-value");
+			this.element.removeAttribute("mv-value");
 			this.parsed = [new Mavo.Expression(expression)];
 			this.expression = expression;
+		}
+
+		if (this.attribute?.startsWith("mv-attr-")) {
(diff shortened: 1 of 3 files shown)


DmitrySharabin:
Here are a couple of examples: https://codepen.io/dmitrysharabin/pen/gOoqOzj

karger:
awkward hack: define another attribute that converts its indicated 
attribute from hyphenated to camel case; e.g. mv-camelcase-view-box 
would become viewBox.

karger:
actually, here is another possible approach: apparently there is a 
specified list of tag adjustments that you could perform automatically 
*if the node with the attributes is an svg node* which I suspect can be 
checked somehow.

https://html.spec.whatwg.org/multipage/parsing.html#adjust-svg-attributes

search for "adjust svg attributes" in this doc if the anchor fails.

LeaVerou:
Oh I'm glad you're working on this.
Is the casing issue cosmetic or does it actually cause loss of functionality?

Due to the way Mavo resolves expressions, substituting gradually, I imagine (without having seen the code) that this solution may still produce errors when the expressions depend on other expressions, e.g.:

```html
<input type=number value="[foo]">
<span property="foo" value="[bar]"></span>
<span property="bar" value="3"></span>
```

DmitrySharabin:
> Is the casing issue cosmetic or does it actually cause loss of functionality?

In the SVG Path Builder we won't be able to get rid of the error caused by this code: `viewBox="0 0 [width] [height]"`.

> Due to the way Mavo resolves expressions, substituting gradually, I imagine (without having seen the code) that this solution may still produce errors when the expressions depend on other expressions...

Do you mean trigger browser errors? The following code works without errors (I updated the pen). :) But I'm almost 100% sure I could miss some use cases.

```html
<input type=number mv-attr-value="[foo]" />
<span property="foo" mv-value="bar"></span>
<span property="bar">3</span>
```

DmitrySharabin:
> actually, here is another possible approach: apparently there is a specified list of tag adjustments that you could perform automatically *if the node with the attributes is an svg node* which I suspect can be checked somehow.

That's interesting. Thanks, David. 🙏🏻

I was thinking of the approach you suggested and wondering: do we really need this at the core? If the casing issue becomes a problem for authors, we can add a plugin that will make the `mv-attr-*` family of attributes case-sensitive. What do you think?

LeaVerou:
> > Is the casing issue cosmetic or does it actually cause loss of functionality?
> 
> In the SVG Path Builder we won't be able to get rid of the error caused by this code: `viewBox="0 0 [width] [height]"`.

Why not? Can't we specify `mv-attr-viewBox` and then your code would specify `viewbox` (lowercase), which I think would still work?

DmitrySharabin:
> Why not? Can't we specify `mv-attr-viewBox` and then your code would specify `viewbox` (lowercase), which I think would still work?

Ah, yes. You are absolutely right—lowercase attributes are not even an issue. They work perfectly fine. Stupid me. 🤦‍♂️

Now I need to figure out why `viewBox`, for example, is not live if it's set via `mv-attr-viewbox`. That's why I thought attributes case is an issue.

Thank you. I'll continue my investigation then.

DmitrySharabin:
Thanks to David, we know that [browsers adjust SVG attributes that are not all lowercase](https://html.spec.whatwg.org/multipage/parsing.html#adjust-svg-attributes). That means that when there is an expression inside the value of an attribute (e.g., inside `viewBox`), Mavo gets the case-sensitive version of the attribute name to work with (e.g., `viewBox`). As a result, `viewBox="0 0 [width] [height]"` works perfectly fine (the view box adjusts according to the values of `width` and `height`).

The `mv-attr-*` attribute, on the other hand, is not the standard SVG attribute. So, for the `mv-attr-viewBox` Mavo will get the `mv-attr-viewbox` ([all lowercase](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#lower_casing)) to work with. And even though SVG understands `viewbox="0 0 [width] [height]"` (when added directly to `<svg>`), if this attribute is added through `mv-attr-viewbox="0 0 [width] [height]"`, the expressions don't have any effect on the SVG. Though, the `viewbox` attribute is being updated via expressions. The same thing happens with `preserveAspectRatio`.

We can conclude that the casing issue causes loss of functionality. So I decided to follow David's suggestion and adjust those kinds of attributes manually. And it worked. Yay!

I updated [the demo](https://codepen.io/dmitrysharabin/pen/gOoqOzj) so that it showcases different use cases.

I believe this PR is ready for review. 🤞🏻

LeaVerou:
Are you using the correct namespace for these attributes? I wonder if that's why `viewbox` doesn't work.

DmitrySharabin:
> Are you using the correct namespace for these attributes? I wonder if that's why `viewbox` doesn't work.

Hmm. That's a good question. I don't know. I definitely need to figure it out.

DmitrySharabin:
Here are some results of my research concerning SVG, its elements, attributes, and namespaces:

1. Regardless of whether we use `viewBox` or `viewbox`, the browser adjusts it to `viewBox` (and adds the corresponding attribute if needed).

Suppose we have (mind the attribute case):
```html
<svg viewbox="0 0 100 100">
  ...
</svg>
```

In DevTools, we see (mind the attribute case):

Screen Shot 2022-04-21 at 19 09 42


After inspecting the `<svg>` element and invoking `console.dir($0)` in the console, we get:

Screen Shot 2022-04-21 at 19 05 19

2. Attributes of SVG elements don't have `namespaceURI` (even though the corresponding elements do). Except `href`, IIRC. See the previous screenshot.

3. Regardless the point 2 (it didn't stop me 😂), I searched through the codebase and found [the method](https://github.com/mavoweb/mavo/blob/5a33b0cb2650bf539eac430dfc069a5f6cd03f31/src/primitive.js#L1069-L1150) where we set attributes: `Primitive.setValue()`. There, I replaced [this line](https://github.com/mavoweb/mavo/blob/5a33b0cb2650bf539eac430dfc069a5f6cd03f31/src/primitive.js#L1133)

```js
element.setAttribute(o.attribute, value);
```

with this one:

```js
if (element.namespaceURI === "http://www.w3.org/2000/svg") {
  element.setAttributeNS(element.namespaceURI, o.attribute, value);
}
else {
  element.setAttribute(o.attribute, value);
}
```

Then I inspected `<svg>` added by the following code (mind the fallback for `viewBox` and its case):

```html
<svg
  viewbox="-50 -50 50 50"
  mv-attr-viewBox="[list(0, 0, 100, 100)]">
  ...
</svg>
```

In DevTools we have

Screen Shot 2022-04-21 at 21 19 01

and

Screen Shot 2022-04-21 at 21 22 27

As we can see, the browser adjusted the fallback attribute (as expected), and the one added by Mavo has the `namespaceURI` property and is evaluated correctly (its value based on an expression), though it doesn't have any effect. It's simply ignored by the browser. Moreover, every other `mv-attr-*` attribute stopped working (even though without namespace they worked fine).

To conclude, `namespaceURI` doesn't have any positive effect in our case.

So, unfortunately I couldn't find the way to make `mv-attr-*` work with case-sensitive SVG attributes without manually tweaking them (as I do it in this PR). 😔 However, I can't deny that there is something I still don't get. 🤷🏻‍♂️

LeaVerou:
I believe `namespaceURI` is null when it can be determined via the parent, ancestor etc namespace. It's not null on `href` because href has the XLink namespace implicitly (in fact it used to be written `xlink:href`). 

I wonder if something like this would work: https://codepen.io/leaverou/pen/ZEvPPaP?editors=1111

DmitrySharabin:
> I wonder if something like this would work: https://codepen.io/leaverou/pen/ZEvPPaP?editors=1111

Whoa! That's cool!!! Thanks for the new knowledge.

Re:context. Got it. Totally agree. I will take that into account.

DmitrySharabin:
Done!

DmitrySharabin:
> One complication is that not all attributes will work on the root element, some of them may need to be on specific elements. You should try and see if it works for e.g. `refX` and that sort of thing.

If I get your comment correctly, now it should work (after some fixes I made to the code): https://codepen.io/dmitrysharabin/pen/mdpgBvy

DmitrySharabin:
@LeaVerou, I would also be grateful if you give this PR another look. I believe the feature it adds to Mavo might be very beneficial. We might make this feature experimental.

src/util.js
```diff
@@ -430,6 +430,14 @@ var _ = $.extend(Mavo, {
 		return element.getAttributeNames().filter(name => regex.test(name));
 	},
 
+	// Fixes the case of attributes that are not all lowercase
+	// Especially useful for SVG attributes
+	// https://html.spec.whatwg.org/multipage/parsing.html#adjust-svg-attributes
+	getProperCasing: (element, attribute, root = "svg") => {
+		let doc = new DOMParser().parseFromString(`<${root}><${element} ${attribute}=""></${element}></${root}>`, "text/html");
```
LeaVerou, line 437: I think parsing is a pretty intensive operation, and these are results that don't change, so we should probably cache them in a structure. 

Then, you can cache the result in a structure like:

```js
{
	"svg": {
		"viewbox": "viewBox", // all-lowercase to proper casing
		...
	},
	"math": {...}
```

src/util.js
```diff
@@ -430,6 +430,14 @@ var _ = $.extend(Mavo, {
 		return element.getAttributeNames().filter(name => regex.test(name));
 	},
 
+	// Fixes the case of attributes that are not all lowercase
+	// Especially useful for SVG attributes
+	// https://html.spec.whatwg.org/multipage/parsing.html#adjust-svg-attributes
+	getProperCasing: (element, attribute, root = "svg") => {
```
LeaVerou, line 436: 1. You don't seem to be using the 3rd argument anywhere. Perhaps a better design might be to pass in an actual element, and get its root via `.closest()` with a hardcoded list of potential roots.
2. This is a global utility method, so `getProperCasing` could mean anything out of context. I'd change to `getTagCase` or `getProperTagCase`.

src/domexpression.js
```diff
@@ -49,7 +60,14 @@ var _ = Mavo.DOMExpression = $.Class({
 		if (typeof this.expression !== "string") { // Still unhandled?
 			if (this.attribute) {
 				// Some web components (e.g. AFrame) hijack getAttribute()
```
LeaVerou, line 62: Need to move this comment to still precede the `Element.prototype.getAttribute` call, otherwise it makes no sense. 
You could also alias it to a variable outside everything, have the comment right above that and use that in calls:

```js
// Some web components (e.g. AFrame) hijack getAttribute()
const originalGetAttribute = Element.prototype.getAttribute;
```

LeaVerou (changes requested):
Looks good overall, I left a few minor comments.

Btw what happens if someone uses *both* `mv-attr-foo` and `foo`? Does the order matter?

DmitrySharabin:
> Btw what happens if someone uses both `mv-attr-foo` and `foo`? Does the order matter?

As you mentioned in your original issue, “This has the added benefit of allowing fallbacks by just providing a value on the actual attribute.” So, yes, the order matters, and `mv-attr-foo` has precedence over `foo` as soon as Mavo steps into play.
E.g,. this doesn't error anymore: `<rect width="40" height="30" rx="3" ry="3" mv-attr-rx="[rx]" mv-attr-ry="[ry]" />`. 🥳

src/domexpression.js
DmitrySharabin: Done! Thanks.

src/util.js
DmitrySharabin: Done! Thank you.

DmitrySharabin:
@LeaVerou, thank you so much for all your comments and suggestions. I took them into account. It would be nice if you could find time to review it again.

LeaVerou:
> > Btw what happens if someone uses both `mv-attr-foo` and `foo`? Does the order matter?
> 
> As you mentioned in your [original issue](https://github.com/mavoweb/mavo/issues/475), “This has the added benefit of allowing fallbacks by just providing a value on the actual attribute.” So, yes, the order matters, and `mv-attr-foo` has precedence over `foo` as soon as Mavo steps into play. E.g,. this doesn't error anymore: `<rect width="40" height="30" rx="3" ry="3" mv-attr-rx="[rx]" mv-attr-ry="[ry]" />`. 🥳

Right, but what happens if the attribute also has an expression?

src/domexpression.js
```diff
@@ -49,7 +60,16 @@ var _ = Mavo.DOMExpression = $.Class({
...
-				var value = Element.prototype.getAttribute.call(this.node, this.attribute);
+				const originalGetAttribute = Element.prototype.getAttribute;
```
LeaVerou, line 63: Since this never changes, you can move it outside the class altogether. (Not a blocker for merging this PR)

src/util.js
```diff
@@ -430,6 +430,35 @@ var _ = $.extend(Mavo, {
...
+	// E.g., { "svg": { "viewbox": "viewBox", ... }, "math": { ... } }
+	properlyCasedAttributesCache: {},
+
+	// Fixes the case of attributes that are not all-lowercase
+	// Especially useful for SVG attributes
+	// https://html.spec.whatwg.org/multipage/parsing.html#adjust-svg-attributes
+	getProperAttributeCase (element, attribute) {
+		const roots = "svg, math"; // Potential root elements
+
+		const root = element.closest(roots).tagName;
```
LeaVerou, line 443: This will return `null` if the element is not within `<svg>` or `<math>` and error. We could finish the selector with `:root` so that it always returns something, or use `?.` (depending on whether this is useful at all beyond SVG/MathML or not).

LeaVerou (changes requested):

src/util.js
DmitrySharabin: I would follow the first route (with the `:root`), even though this utility is only used with SVG/MathML for now.

Thank you for pointing this out.

src/domexpression.js
DmitrySharabin: Done! Thanks.

LeaVerou:
I think as long as it's deterministic, there are many right choices. The main thing we should avoid is it being non-deterministic, i.e. *sometimes* it overrides it and sometimes it doesn't depending on race conditions.

I agree `mv-attr-foo` always having precedence over `foo` makes sense. I could also be persuaded that the order of attributes can specify precedence.

Thoughts, @karger @ adamjanicki2 ?

adamjanicki2:
> I think as long as it's deterministic, there are many right choices. The main thing we should avoid is it being non-deterministic, i.e. _sometimes_ it overrides it and sometimes it doesn't depending on race conditions.
> 
> I agree `mv-attr-foo` always having precedence over `foo` makes sense. I could also be persuaded that the order of attributes can specify precedence.
> 
> Thoughts, @karger @adamjanicki2 ?

I think overall I like having `mv-attr-attrname` taking precedence or `attrname`, but also there is some precedence with ordering in other frameworks, where typically the last attr specified takes precedence (for example in react if you have something like `<Element {...props} className="..." />`), so I could also be persuaded for an ordering, but since Mavo has this specific feature for defining attributes, the goal should be encouraging users to use them over the default ones, which is why I like Mavo attrs taking precedence

Regardless of what we decide, we definitely will need to update the documentation to be very clear and explicit about how this works

src/domexpression.js
```diff
@@ -31,6 +34,17 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.expression = expression;
 		}
 
+		if (this.attribute?.startsWith("mv-attr-")) {
```
adamjanicki2, line 37: It would be great to refactor this into a helper function since it's used multple times in this file, e.g.:

```
const MAVO_ATTR_PREFIX = 'mv-attr-';

function isMavoAttribute(attribute) {
     return attribute.startsWith(MAVO_ATTR_PREFIX);
}
```

LeaVerou: Might be worth having a constant with the prefix (marginally, so I'll leave it up to @DmitrySharabin ), but having a helper function to avoid two `startsWith()` calls is the definition of [premature abstraction](https://medium.com/@thisdotmedia/the-cost-of-premature-abstraction-b5d71ffd6400). In general, [duplication is cheaper than the wrong abstraction](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction), so I tend to follow (and recommend) the [rule of three strikes](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

DmitrySharabin: Honestly, I couldn’t find a good place for a constant holding the attribute name, so I decided to stick with the approach we already have for now. Moreover, it fits nicely with the rest of the code. If someday we decide to change the attribute name and refactor the code, I believe Find/Replace will quickly get the job done. 😄

src/domexpression.js
```diff
@@ -120,7 +120,17 @@ var _ = Mavo.DOMExpression = $.Class({
...
+		if (expressions.some(e => e.originalAttribute?.startsWith("mv-attr-"))) {
+			for (let expression of expressions.filter(e => !e.originalAttribute?.startsWith("mv-attr-"))) {
+				// We already have expressions associated with this attribute; discard them since mv-attr-* has priority
+				expression.active = false;
+			}
+		}
+
+		if (this.active !== false) {
+			this.mavo.expressions.register(this);
+		}
```
DmitrySharabin, line 133: @LeaVerou, @adamjanicki2, I came up with this way of being deterministic when both `foo` and `mv-attr-foo` have an expression, and we expect `mv-attr-foo` to take precedence. Could you please review it?

LeaVerou: What happened before? Did it follow the order of attributes or was it non-deterministic?

DmitrySharabin: As I learned from my experiments, it did follow the order of attributes. However, I can’t guarantee that it was deterministic all the time. I believe some inconsistency might appear because of race conditions. But I’m not sure. That’s why I’m leaning towards the `mv-attr-*` wins route.

LeaVerou: I think a better route would be to prevent the creation of these useless objects altogether, rather than having to make them inactive later. I think a better place for this logic might be in `extract()` or `traverse()` in [expressions.js](https://github.com/mavoweb/mavo/blob/master/src/expressions.js), especially since that's where other conditions for ignoring attributes are handled (e.g. `mv-expressions-ignore`).

DmitrySharabin: Good idea. I struggled to find an appropriate place for the logic I meant to implement—I tried to prevent the creation of the expressions we didn't mean to use but without luck. I will commit another try, but in `expressions.js` this time. Thank you!

src/expressions.js
```diff
@@ -223,11 +223,26 @@ var _ = Mavo.Expressions = $.Class({
...
+				let ignorable = attributes.find(attribute => attribute.name === name);
+
+				if (ignorable) {
+					// Ignore expressions in foo, if any, since mv-attr-foo has priority
+					ignore ??= new Set();
+					ignore.add(ignorable.name);
+				}
+			}
+
+			attributes.forEach(attribute => {
```
DmitrySharabin, line 245: @LeaVerou, thank you so much for your hint—`expressions.js` is indeed the appropriate place for this logic. Sorry for requesting your review one more time. 🙈

LeaVerou: I think this may be complicating things a bit. I just pushed a commit that tries to simplify the logic.

Also, we don't need to apply `syntax.test()` here, the values of `mv-attr-*`, like directives, are *always* expressions. If you haven't designed it like that, you should.

DmitrySharabin: I re-designed the `mv-attr-*` feature to work as `mv-value`. Because of the wrong design decision, I wrote way too complicated logic. I knew that `mv-attr-foo` doesn't make sense if it's not an expression, but I chose the wrong path to implement it. Thank you for pointing this out. 🙏🏻

src/domexpression.js
```diff
@@ -31,6 +34,21 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.expression = expression;
 		}
 
+		if (this.attribute?.startsWith("mv-attr-")) {
+			this.originalAttribute = this.attribute;
+			this.attribute = this.attribute.replace("mv-attr-", "");
+
+			if (["http://www.w3.org/2000/svg", "http://www.w3.org/1998/Math/MathML"].includes(this.element.namespaceURI)) {
```
LeaVerou, line 41: Nitpick but we should move these URLs to constants (e.g. `SVG_NAMESPACE_URI` and `MATHML_NAMESPACE_URI`)

src/domexpression.js
```diff
@@ -31,6 +34,21 @@ var _ = Mavo.DOMExpression = $.Class({
...
+				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
+			}
+
+			this.fallback = this.fallback || Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
```
LeaVerou, line 45:
```suggestion
			this.fallback ||= Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
```

(also do we need `||` or `??` here?)

src/expressions.js
```diff
@@ -183,11 +183,11 @@ var _ = Mavo.Expressions = $.Class({
 	},
 
 	extract: function(node, attribute, path, syntax = Mavo.Expression.Syntax.default) {
-		if (attribute && _.skip.indexOf(attribute.name) > -1) {
+		if (_.skip.includes(attribute?.name)) {
```
LeaVerou, line 186: Can we remove this entirely now that `_.skip` is taken into account in `traverse()`? As far as our own codebase is concerned, that seems ok, but not sure if any plugins call `extract()` and depend on this behavior

src/expressions.js
```diff
@@ -183,11 +183,11 @@ var _ = Mavo.Expressions = $.Class({
...
 			return;
 		}
 
-		if (attribute && _.directives.indexOf(attribute.name) > -1 ||
+		if (_.directives.includes(attribute?.name) || attribute?.name?.startsWith("mv-attr-") ||
```
LeaVerou, line 190: I wonder if we can do something less ad hoc here. E.g. make `_.directives` a list of strings and regexes, then check like:

```js
let attributeName = attribute?.name;
if (_.directives.some(d => d.test?.(attributeName) || d === attributeName)) || ...
```

src/domexpression.js
DmitrySharabin: I assume that some falsy values (e.g., `0`) might be perfectly valid fallbacks, and we don't want to replace them with any other value. So, I'm leaning towards `??=`.

src/domexpression.js
DmitrySharabin: Agreed. Done!

src/expressions.js
DmitrySharabin: Good question. I searched for `extract(` in our repo with plugins and found no mentions. Neither could I find this in my repos. I'm unsure if we have any other plugins we might break. We can try to commit to this change and see what might happen. 🤓

src/expressions.js
DmitrySharabin: Great idea! Done.

src/expressions.js
DmitrySharabin: I'm ready to remove it as soon as we decide to do that. I commented this block out and everything still works.

LeaVerou: Let's do it!

DmitrySharabin: Done!

Now, I'm ready to request your review one more time. Thanks a ton for your patience and excellent ideas! 🙏🏻

src/domexpression.js
```diff
@@ -31,6 +37,21 @@ var _ = Mavo.DOMExpression = $.Class({
 			this.expression = expression;
 		}
 
+		if (this.attribute?.startsWith("mv-attr-")) {
+			this.originalAttribute = this.attribute;
+			this.attribute = this.attribute.replace("mv-attr-", "");
+
+			if ([SVG_NAMESPACE_URI, MATHML_NAMESPACE_URI].includes(this.element.namespaceURI)) {
+				this.attribute = Mavo.getProperAttributeCase(this.element, this.attribute);
```
LeaVerou, line 45: Just dawned on me: do we also need to do this in XHTML documents (with an actual XML MIME type)? If so, it would require a different check.

src/domexpression.js
```diff
@@ -31,6 +37,21 @@ var _ = Mavo.DOMExpression = $.Class({
...
+			}
+
+			this.fallback ??= Mavo.Primitive.getValue(this.element, {attribute: this.attribute});
+			let expression = this.element.getAttribute(this.originalAttribute);
+			this.element.removeAttribute(this.originalAttribute);
```
LeaVerou, line 50: There seems to be a lot of repetition between this and the preceding block implementing `mv-value`. I'll push a commit to address it.

DmitrySharabin: Whoa! Cool! It's working with this beautiful piece of code. 🥳 Thank you!!

src/domexpression.js
DmitrySharabin: Probably we should. Our users will push us to do this in any case. 😅

LeaVerou (approved):
Looks like this is done then?

DmitrySharabin:
I think so! 🤞

DmitrySharabin:
@LeaVerou, thanks a ton for your help! We made this awesome feature happen. Yay! 🥳

karger:
> I think as long as it's deterministic, there are many right choices. The main thing we should avoid is it being non-deterministic, i.e. _sometimes_ it overrides it and sometimes it doesn't depending on race conditions.
> 
> I agree `mv-attr-foo` always having precedence over `foo` makes sense. I could also be persuaded that the order of attributes can specify precedence.
> 
> Thoughts, @karger @adamjanicki2 ?

I think the natural use case for foo= mv-attr-foo= is when you want to give the browser something to work with until the expression in mv-attr-foo is evaluated, but that once that happens mv-attr-foo should dominate.

DmitrySharabin:
> I think the natural use case for foo= mv-attr-foo= is when you want to give the browser something to work with until the expression in mv-attr-foo is evaluated, but that once that happens mv-attr-foo should dominate.

This is exactly how it’s designed and implemented. I’m glad we are all on the same page. ☺️

karger:
II should have been clearer.  I was responding to the previous post, and saying that I think mv-attr-foo should dominate foo instead of relying on order of attributes.

> > I think the natural use case for foo= mv-attr-foo= is when you want to give the browser something to work with until the expression in mv-attr-foo is evaluated, but that once that happens mv-attr-foo should dominate.
> 
> This is exactly how it’s designed and implemented. I’m glad we are all on the same page. ☺️

karger:
> > I think the natural use case for foo= mv-attr-foo= is when you want to give the browser something to work with until the expression in mv-attr-foo is evaluated, but that once that happens mv-attr-foo should dominate.
> 
> This is exactly how it’s designed and implemented. I’m glad we are all on the same page. ☺️

Having considered further, I can see an argument for order of attributes, because that's how html works, right?   So once the expression is evaluated. we could apply the standard order of attributes rules to decide if it overrides the foo attribute.

DmitrySharabin:
@karger, do you mind filing a separate issue so we can discuss it there, not in the merged PR? There is always room for improvements we don't want to skip. :)

PR #877, opened 2022-07-02
Attempt to fix #876: Do not descend at all in objects that are not plain objects

### LINKED ISSUE #876: Do not mess with objects in data that are not plain objects
When rendering data onto an app, instances of some class other than Object should be left alone. Instead, Mavo adds proxies and all sorts of crap on these values, breaking all sorts of code. This is why we were having to do all these awkward hacks with the Color.js demos.

Testcase: https://codepen.io/leaverou/pen/QWmbpmm

LeaVerou:
Hi @DmitrySharabin, could you please check if this introduces regressions? Thanks!

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -64,9 +64,6 @@ var _ = Mavo.Data = $.Class(class Data {
 				// Why check prototype instead of just type == "object"? Because instances of ES6 classes also return "object"
 				_.computeRoutes(this.data);
 			}
-			else {
-				_.computeMetadata(this.data, this.key, this.parent);
-			}
 
 			this.updateParent();
 		}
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
Hey @LeaVerou, with the help of Requestly (redirecting all the requests of `mavo.js` to https://deploy-preview-877--getmavo.netlify.app/mavo.js), I checked: 

- all the tests we have
- the website
- and all the demos (both on the website and on CodePen) we have (including some of from the _Mavo in the wild_ list)
- all of my apps
- and a couple or yours: the Forkgasm website, the color pickers, MaVoice, the Home search app, the Issue closing stats app, and the TAG breakouts app (on CodePen).

I found no new regressions. It seems to me, everything works just fine.

PR #913, opened 2022-09-01
Fix the output for ordinal numbers, closes #912

### LINKED ISSUE #912: Let the pluralize() function return `${num}${label}` (without the space) when type: "ordinal" is provided
It will let us get the correct result (`42nd`) for `pluralize(42, one: "st", two: "nd", few: "rd", other: "th", type: "ordinal")` instead of `42 nd` we get now.

For now, to get the correct result we should do something similar to this:
`<p>42[pluralize(42, one: "st", two: "nd", few: "rd", other: "th", type: "ordinal", text_only: true)]</p>`.

I believe in languages other than English, ordinal indicators are not divided by space from the preceding number as well.

DmitrySharabin:


LeaVerou (approved):

LeaVerou:
Thank you! 🙏🏼

PR #923, opened 2022-09-13
Fix for the `intersection()` function

DmitrySharabin:


LeaVerou (changes requested):
Dropping proxies fixes it, but also means we lose the reference back to the property the items came from. We should use `val()` when comparing items (ideally with some way to make sure we don't do this multiple times per item).

DmitrySharabin:
> Dropping proxies fixes it, but also means we lose the reference back to the property the items came from.

Thank you very much for pointing this out. I haven't thought about it in advance. I tried another approach in the new commit. Please have a look when you have a spare cycle.

lewisushindi (commented):
looks fine according to me

LeaVerou (approved):
LGTM!

PR #929, opened 2022-10-08
Attempt to fix #741

### LINKED ISSUE #741: @mv-default with an expression on <select> appends an extra <option> with this expression
Testcase: https://codepen.io/dmitrysharabin/pen/vYmOYdg?editors=1000

Git bisect showed this commit as the first bad one: 6acb4088d3f538b81b59bb7bde30173ff6f6f62b.

Interestingly, [this issue](https://github.com/mavoweb/mavo/issues/716) found by @karger is not the case anymore.

DmitrySharabin:


src/primitive.js
```diff
@@ -183,7 +183,7 @@ var _ = Mavo.Primitive = class Primitive extends Mavo.Node {
 		if (this.editor) {
 			if (_.isFormControl(this.editor)) {
 				if (this.editor.matches("select")) {
-					let text = [...this.editor.options].find(o => o.value == value)?.textContent;
+					let text = [...this.editor.options].find(o => (Array.isArray(value)? value : [value]).map(v => v + "").includes(o.value))?.textContent;
```
LeaVerou, line 186: Isn't `Array.isArray(value)? value : [value]` basically `Mavo.toArray()`? (IIRC)

`v + ""` will fail badly if array items are objects, is that a possibility?

DmitrySharabin: > Isn't Array.isArray(value)? value : [value] basically Mavo.toArray()? (IIRC)

You are absolutely right! Thanks. Somehow I forgot about this utility method.

DmitrySharabin: > v + "" will fail badly if array items are objects, is that a possibility?

Hm. I don’t know. I need to investigate it.

DmitrySharabin: > v + "" will fail badly if array items are objects

Will the `toString()` method do in that case? It seems to me, it covers all the cases we might face.

LeaVerou: No, `toString()` does exactly the same as `v + ""`. E.g. `({foo: 2}).toString()` is `'[object Object]'`, which is not really helpful.

LeaVerou (approved):

LeaVerou:
Better late than never! :D

Issue #925, opened 2022-09-29
Negative values in duration() should produce a more meaningful result than "0 ms"


LeaVerou:
Options:

- We just take the absolute value and ignore that it's negative. So `duration(-10000000)` and `duration(10000000)` would both return `"2 hours"`. 
- We could return a negative duration (e.g. `"-2 hours") `to indicate to the user that there's a mistake, without breaking their UI too much if they're not there to debug

Note that we cannot do things like `"2 hours ago"` because that wouldn't work with localization.

And while I'm at it, `"0 ms"` is almost never desirable for `duration(0)`. We should return something else, though not sure what.

DmitrySharabin:
> Note that we cannot do things like "2 hours ago" because that wouldn't work with localization.

Actually, this way of handling negative values makes the most sense from my perspective. That was my first thought when I saw the issue title, to be honest. 😀 Would it be possible to use [this platform API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) to handle the localization issue you mentioned?

It seems to me, that can also cover the `duration(0)` case.

LeaVerou:
Oh that would be really cool. We can do that when the argument is negative and keep the current behavior when it's positive.
That does make it a little more intensive to implement though. Are you interested in implementing it @DmitrySharabin? Or @barishnamazov since you were the one that pointed this out, perhaps implementing that would be of interest?

BarishNamazov:
@DmitrySharabin's idea is really great, but I was wondering if this breaks the consistency of duration's usage. For positive values, `duration` returns an amount of time, but if we return `2 hours ago` for negative values, the result would be a point in time. Is it okay that the units don't match? If the code is like `I slept ${duration(end_time - start_time)}` or `Sleep duration: ${duration(end_time - start_time)}`, positive values definitely makes sense, but negative values would be very confusing (first example is especially interesting, changing the meaning of the sentence). I would personally prefer seeing a negative amount of time (e.g. `-2 hours`). I think what would work the best for duration(0) would be `0 unit` where unit is provided (if not, I think we can use hours).

Let me know what you think!

DmitrySharabin:
Yes, I also see that inconsistency. So I agree.

What if we follow option 2 (with the negative value as the result) and provide a `relative_time()` (or whatever we will call it) function for those who need it? It will cover a bunch of other use cases with relative time (not only in the past, but in the future as well).

What do you think?

[...]

PR #930, opened 2022-10-08
Negative inputs to `duration` now produce negative time outputs

BarishNamazov:
Solves the main issue agreed on: https://github.com/mavoweb/mavo/issues/925
For 0 input, still returns `0 ms`.

### DIFF
--- a/src/functions.date.js
+++ b/src/functions.date.js
@@ -234,6 +234,10 @@ _.duration = $.extend(function (ms, terms) {
 	// TODO unify code for specific unit with code for auto units to reduce repetition
 	// TODO allow multiple units, e.g. ["days", "hours"]
 	// TODO allow combining term # and units, e.g. start: days, terms: 2
+	
+	let negativeMultiplier = ms < 0 ? -1 : 1; // a multiplier to convert result to negative if needed
+	ms = Math.abs(ms); // negative works same way as positive does, just adding negative sign in the front
+
 	if (terms && isNaN(terms)) {
 		// Specific term specified
 		let unitSingular = terms != "ms" ? terms.replace(/s?$/, "") : terms;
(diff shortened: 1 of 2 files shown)


LeaVerou (approved):
LGTM. 

Note that `===` will not work if `ms` comes directly from a property (as it will be an object), though I see it's used above too, so I'm not gonna block this for that, but `==` would work for both cases. 

Before we merge, could you add some tests too? This is the tests repo: https://github.com/mavoweb/test

BarishNamazov:
Good call! There's something interesting with negative numbers I just noticed (I had the feeling of breaking some math rules while implementing this, but just couldn't get my head around it back then). With positive numbers, we round it down (e.g., `2 hours 51 minutes` become `2 hours`), but rounding down mathematically works differently for negative numbers (`-2 hours 5 minutes` become `-3 hours`). Which one do you think makes more sense?

I don't have a strong preference since it's rounding anyway, but I feel like to keep it consistent with positive numbers (though not mathematically), like `-2 hours 51 minutes` becoming `-2 hours` makes more sense to me. I caught this when I was writing tests:

image

(note that this is the part of the implementation that happens in the first if statement in the function, not affected by this PR)

What do you think? Should I do that change and push it here? Or we should stick to the mathematical version (both are easy to implement: doing it like positives needs you to convert to abs value and then convert back to negative, while other one doesn't require any special treatment)?

LeaVerou:
I agree with your intuition: the rounding should operate on the absolute value, instead of doing proper mathematical rounding, since I suspect in most cases this will just be caused by people subtracting in the wrong order so it will be confusing if they get a different result once they fix the order.

BarishNamazov:
should be done now!

LeaVerou:
Great work, merged! 
In the future, please include phrases like "closes #925" in one of your commit messages so that it automatically closes the relevant issue, see https://github.blog/2013-01-22-closing-issues-via-commit-messages/

PR #932, opened 2022-10-09
Fix #661 change requestAnimationFrame to setTimeout

### LINKED ISSUE #661: Expression evaluation does not happen when tab is inactive
We use `requestAnimationFrame()` for each computation cycle, which as it turns out [is paused in inactive tabs](https://stackoverflow.com/questions/15871942/how-do-browsers-pause-change-javascript-when-tab-or-window-is-not-active). We should switch to using `setTimeout()` with an appropriate timeout. That will also be slowed to 1s in inactive tabs, but at least it won't be paused completely.

We should do a code review and replace any other critical rAFs as well.

daclouds:
I don't know if I understand the issue properly.
If I misunderstood, please leave a comment.

### DIFF
--- a/src/expressions.js
+++ b/src/expressions.js
@@ -75,7 +75,9 @@ var _ = Mavo.Expressions = $.Class({
 			}
 		}
 		else {
-			requestAnimationFrame(() => this.update(evt));
+			setTimeout(() => {
+				this.update(evt);
+			}, _.THROTTLE);
 		}
 	},
 


LeaVerou:
Thanks! Looks good, not sure about the timeout, I suspect it may need to be shorter than the overall throttle. Have you tested that tests pass and existing demos work with this change? @DmitrySharabin you might have something to suggest here.

PR #940, opened 2022-12-13
Added underscore to function fromLast()

ghost:


LeaVerou (changes requested):
Hi there,
Thank you for your contribution.
Please note we never modify old versions. They are there for archival purposes, to be served by `get.mavo.io`. Modifying old versions defeats the entire point of versioning.

Issue #915, opened 2022-09-03
`between()` produces an incorrect result when the `from` parameter is more than one character long


DmitrySharabin:
And since some other functions (like `from()` and `fromlast()`) use it under the hood, all of them produce incorrect results in such cases as well.

Testcase: https://codepen.io/dmitrysharabin/pen/ZEoENPM

PR #946, opened 2023-01-12
fixed between function

mohammadKhalafi:
issue link : https://github.com/mavoweb/mavo/issues/915

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -609,8 +609,8 @@ var _ = Mavo.Functions = {
 	between: $.extend((haystack, from, to, tight) => {
 		[haystack, from, to] = [str(haystack), str(from), str(to)];
 
-        	let fromIndex = from ? haystack[tight? "lastIndexOf" : "indexOf"](from) : 0;
-		let toIndex = to ? haystack[tight? "indexOf" : "lastIndexOf"](to) : haystack.length;
+		let fromIndex = from? haystack[tight? "lastIndexOf" : "indexOf"](from) : 0;
+		let toIndex = to? haystack[tight? "indexOf" : "lastIndexOf"](to) : haystack.length;
 
 		if (fromIndex === -1 || toIndex === -1) {
 			return "";
@@ -621,7 +621,6 @@ var _ = Mavo.Functions = {
 		}
 
 		return haystack.slice(fromIndex + from.length, toIndex);
-		
 	}, {
 		multiValued: true
 	}),
--- a/src/functions.js
+++ b/src/functions.js
@@ -609,14 +609,18 @@ var _ = Mavo.Functions = {
 	between: $.extend((haystack, from, to, tight) => {
 		[haystack, from, to] = [str(haystack), str(from), str(to)];
 
-		const fromIndex = haystack[tight? "lastIndexOf" : "indexOf"](from);
-		const toIndex = haystack[tight? "indexOf" : "lastIndexOf"](to);
+		let fromIndex = haystack[tight? "lastIndexOf" : "indexOf"](from);
+		let toIndex = haystack[tight? "indexOf" : "lastIndexOf"](to);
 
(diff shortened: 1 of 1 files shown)


src/functions.js
```diff
@@ -609,14 +609,14 @@ var _ = Mavo.Functions = {
...
+		const fromIndex = haystack[tight? "lastIndexOf" : "indexOf"](from);
+		const toIndex = haystack[tight? "indexOf" : "lastIndexOf"](to);
 
-		if (from && i1 === -1 || i2 === -1) {
+		if (fromIndex === -1 || toIndex === -1) {
 			return "";
 		}
 
-		return haystack.slice(i1 + 1, i2 === -1 || !to? haystack.length : i2);
+		return haystack.slice(from ? fromIndex + from.length : 0, to ? toIndex : haystack.length);
```
DmitrySharabin, line 619: If we provide the `tight` parameter equals `true`, won't `fromIndex` become larger than `toIndex`? It looks like in that case we'll get an empty string in _all cases_.

For example, `between("betweentne", "t", "e", true)` will return an empty string instead of `tween` ([according to the docs](https://mavo.io/docs/functions/#between)).

I updated [the pen](https://codepen.io/dmitrysharabin/pen/ZEoENPM) to reflect this case.

DmitrySharabin (changes requested):
Hey there @mohammadKhalafi,

Thank you so much for your interest in Mavo and your PR. It indeed fixes the bug, and [all the tests we have](https://test.mavo.io/functions.html#string-slicing) still pass. That's awesome! 

However, I found a use case where the between function doesn't work as we expect (see the comment). Is there a chance you could fix that issue as well?

src/functions.js
```diff
@@ -609,14 +609,18 @@ var _ = Mavo.Functions = {
...
+		if (fromIndex === -1 || toIndex === -1) {
 			return "";
 		}
 
-		return haystack.slice(i1 + 1, i2 === -1 || !to? haystack.length : i2);
+		if (tight && toIndex <= fromIndex){
+			return haystack.slice(toIndex + to.length, fromIndex);
+		}
+
+		return haystack.slice(fromIndex + from.length, toIndex);
```
DmitrySharabin, line 623: But what if `from` and/or `to` are not provided? They are _optional_ by design (and you took it into account in your first commit). Moreover, the `fromLast()` and `toFirst()` functions are built on top of the optionality of `from` and `to`.

E.g., With this change, `fromlast("foo.bar.baz", ".")` returns `foo.bar` instead of `baz`, and `tofirst("foo.bar.baz", ".")` returns `bar.baz` instead of `foo`.

DmitrySharabin (changes requested):

src/functions.js
mohammadKhalafi: fixed!

DmitrySharabin: Awesome! LGTM. Thanks!

DmitrySharabin (approved):

DmitrySharabin:
@LeaVerou Lea, I reviewed this PR. And it looks like now everything works as it should: the tests pass, the bug is fixed, I also checked our demos—no regression. Do you mind me merging this PR?

mohammadKhalafi:
>

LeaVerou:
> @LeaVerou Lea, I reviewed this PR. And it looks like now everything works as it should: the tests pass, the bug is fixed, I also checked our demos—no regression. Do you mind me merging this PR?

If I assign you a review, it means I trust you to merge when you think it's ready 😄

DmitrySharabin:
> If I assign you a review, it means I trust you to merge when you think it's ready 😄 

Got it. Thanks! 😄

DmitrySharabin:
@mohammadKhalafi Thank you so much for your work!

PR #949, opened 2023-01-21
Insert Mavo bar in `<body>` if Mavo root is on `<html>` fixes #830

### LINKED ISSUE #830: Mavo bar should be inserted in <body> when mv-app is on <html>
There is literally *no* use case where authors would want the Mavo bar to be inserted as a direct child of `<html>` before `<head>` and even if there were, they can override the default placement by using their own element.

rafidoth:
- Mavo bar will insert only when the tagName = "HTML"
- Replaced code in all of the files where the Mavo bar is inserted.

Let me know If there is any mistake in my contribution as I'm new to open-source contributions. Thank you.

### DIFF
--- a/dist/0.1.3/mavo.es5.js
+++ b/dist/0.1.3/mavo.es5.js
@@ -2129,7 +2129,7 @@ function _toConsumableArray(arr) { if (Array.isArray(arr)) { for (var i = 0, arr
 			} else {
 				this.element = $.create({
 					className: "mv-bar mv-ui",
-					start: this.mavo.element.tagName === "HTML"? document.body : this.mavo.element,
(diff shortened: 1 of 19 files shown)


LeaVerou (changes requested):
Hi there, thanks for contributing!

Please do not modify old versions in `dist/`. These are archival and should never be modified. 

@DmitrySharabin can we update `CONTRIBUTING.md` to mention this? It's the 2nd time this happened.

Wrt the actual code change, LGTM.

rafidoth:
Sorry, I didn't notice. It was my bad. Just Updated the present version.

LeaVerou:
Merged, thanks!

Issue #353, opened 2018-06-18
Change mv-github-commit-prefix to mv-github-commit-message


LeaVerou:
There is absolutely no reason not to set the whole message, since it's pretty much the same across commits ("Updated foo.json").

Mayank1728:
hey @LeaVerou  !  I would like to contribute

LeaVerou:
> There is absolutely no reason not to set the whole message, since it's pretty much the same across commits ("Updated foo.json").

Although it *is* different the first time, if the file doesn't already exist ("Created foo.json"). So maybe they should co-exist: if `mv-commit-message` is set, use that as the whole message, if `mv-commit-prefix` is set, use that as a prefix, and if they're both set, I guess prefix the former with the latter.

TusharP05:
Please assign this issue to me if anyone is not Woking on it!

PR #950, opened 2023-01-21
Changed mv-github-commit-prefix to mv-github-commit-message #353

rafidoth:
Changed mv-github-commit-prefix to mv-github-commit-message #353
- Updated backend.github.js

### DIFF
--- a/src/backend.github.js
+++ b/src/backend.github.js
@@ -186,7 +186,7 @@ let _ = Mavo.Backend.register(class Github extends Mavo.Backend {
 
 		let repoCall = `repos/${this.username}/${this.repo}`;
 		let fileCall = `${repoCall}/contents/${path}`;
-		let commitPrefix = this.mavo.element.getAttribute("mv-github-commit-prefix") || "";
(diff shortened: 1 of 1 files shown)


LeaVerou (changes requested):
Hi there, thanks so much for contributing. However, please note that #353 was just a proposal, that had not been discussed with anyone else. Looking back, I'm not sure it was a good idea, for the reasons I described in https://github.com/mavoweb/mavo/issues/353#issuecomment-1399316159
Also, at this point removing `mv-commit-prefix` would be backwards incompatible.
Furthermore, your code just renames the attribute, the whole point was that the `mv-commt-message` would set the **entire** message, not just a prefix. So the code where this is used would need to change too (and `commitPrefix` would also not be a good name).

I'm gonna go ahead and close this, feel free to open a new one if you want to work on a new version, though I'd suggest waiting for some discussion to happen in #353 first. I'd want at least @DmitrySharabin or @karger to weigh in as well.

Thanks again!

rafidoth:
Thank you so much for the simple and kind explanation.

flotaro:
@LeaVerou I would like to contribute, can i do It?

PR #957, opened 2023-03-11
Fix the subtract operator when the first argument is with (unary) minus

DmitrySharabin:
It feels like something crucial to fix before the next release.

@LeaVerou, could you please take a look?


LeaVerou:
LGTM!

PR #960, opened 2023-04-15
Mavoscript functions names change

Ironman0505:
just changed the 2 function names which i have found without underscores. Hope that helps :)


Ironman0505:
If there are any issues, please let me know.

PR #968, opened 2023-09-15
Do not try to load data if we are already doing it, fixes #910

### LINKED ISSUE #910: `mv-login` fired during the data loading breaks the autosave feature
Reduced testcase: https://codepen.io/dmitrysharabin/pen/ZExZaOJ

DmitrySharabin:


LeaVerou (changes requested):
Doesn't `inProgress` contain UI text for *any* kind of in-progress event, and not just data loading? Perhaps there's a conditional that will *only* match when data is being loaded?

DmitrySharabin:
> Doesn't `inProgress` contain UI text for _any_ kind of in-progress event, and not just data loading? Perhaps there's a conditional that will _only_ match when data is being loaded?

You are absolutely right; we have such a conditional. Previously, I thought there was no need to be as precise as to handle the loading in-progress event only. Now, I believe that we should. I have committed the fix. Thank you!

src/mavo.js
```diff
@@ -135,7 +135,7 @@ let _ = self.Mavo = $.Class(class Mavo {
 		$.bind(this.element, "mv-login.mavo", evt => {
 			if (evt.backend == (this.source || this.storage)) {
 				// If last time we rendered we got nothing, maybe now we'll have better luck?
-				if (!this.root.data && !this.unsavedChanges) {
+				if (this.inProgress !== "Loading" && !this.root.data && !this.unsavedChanges) {
```
LeaVerou, line 138: This is not localized?

DmitrySharabin: For now, the only thing that is localized (internally) is `Uploading` (https://github.com/search?q=repo:mavoweb/mavo%20this.inProgress&type=code), but not everywhere. However, all the in-progress events are localized and are present in `locale.en.js`. Should I submit a separate PR to localize the rest of the events? Or we can have everything done in this one.

DmitrySharabin: Btw, we already discussed in-progress events localization here: https://github.com/mavoweb/mavo/pull/799#issuecomment-983952916

And there, we have some cases where localization might be undesirable.

DmitrySharabin: One more thing: `inProgress` is an internal thing. When we show something to the user, we use the localized strings. So I believe we don't need to localize it in other places.

https://github.com/mavoweb/mavo/blob/286de53ec3276a2828f2488d1aaaa5b6170bc631/src/mavo.js#L926-L930

LeaVerou: We should probably lowercase the values (e.g. `"loading"` instead of `"Loading"` to make it clear this is an enum, and not user facing text), then we'd use these as localization keys for the user facing text.

LeaVerou: And yes, it looks like this should be a separate PR, that is merged before this one.

DmitrySharabin: We'll do it if you don't mind. :)

DmitrySharabin: Done!

LeaVerou (approved):
LGTM, approved, feel free to merge once the other one is merged

PR #970, opened 2023-09-20
Make `inProgress` an enum

DmitrySharabin:
Discussed with @LeaVerou [here](https://github.com/mavoweb/mavo/pull/968#discussion_r1328711808).

### DIFF
--- a/src/mavo.js
+++ b/src/mavo.js
@@ -545,7 +545,7 @@ let _ = self.Mavo = $.Class(class Mavo {
 		this.autoSave = false;
 
 		if (data === undefined) {
-			this.inProgress = "Loading";
+			this.inProgress = "loading";
 
 			await backend.ready;
 
@@ -597,7 +597,7 @@ let _ = self.Mavo = $.Class(class Mavo {
 			return;
 		}
 
-		this.inProgress = "Saving";
+		this.inProgress = "saving";
 
 		let saved;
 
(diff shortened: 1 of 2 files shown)


LeaVerou:
Is this all? Don't we need to change localization keys and/or the code that displays this?

DmitrySharabin:
> Is this all? Don't we need to change localization keys and/or the code that displays this?

Yup, that's all. We already have localization keys that match our code:

https://github.com/mavoweb/mavo/blob/286de53ec3276a2828f2488d1aaaa5b6170bc631/src/locale.en.js#L23-L25

All the rest of the code is perfectly valid with these changes:

https://github.com/mavoweb/mavo/blob/286de53ec3276a2828f2488d1aaaa5b6170bc631/src/mavo.js#L926-L930

https://github.com/mavoweb/mavo/blob/286de53ec3276a2828f2488d1aaaa5b6170bc631/src-css/mavo.scss#L241-L244

The corresponding [tests](https://test.mavo.io/bar.html#progress-indicator-styling) “pass” as well. :)

LeaVerou:
> ```js
> this.element.style.setProperty("--mv-progress-text", value? `"${this._(value)}"` : "");
> ```

Orthogonal to this PR, but yikes, that's an antipattern. We may want to do something like `JSON.stringify(this._(value))` instead, to handle escaping quotes.

LeaVerou (approved):

PR #971, opened 2023-09-23
[collection.js] Fixes

DmitrySharabin:
@LeaVerou, it seems to me I could fix both issues we have with collections: `mv-accepts` and drag & drop. 🤞🏻 Could you please have a look?


LeaVerou (approved):
Great work!

PR #972, opened 2023-09-26
Make mv-options work with expressions, fixes #777

### LINKED ISSUE #777: mv-options doesn't work correctly with expressions
Testcase: https://codepen.io/leaverou/pen/bGrVwzM

In the pen it only seems to happen in read mode, but in my app I've definitely also seen this with `mv-mode="edit"`, though I have not isolated the behavior yet.

DmitrySharabin:
I thought it would be nice if such an awesome feature like `mv-options` would work with expressions correctly in the upcoming release. :)

### DIFF
--- a/src/primitive.js
+++ b/src/primitive.js
@@ -1316,6 +1316,9 @@ Mavo.observe({id: "primitive"}, function({node, type, attribute, record, element
 		else if (attribute === "mv-options") {
 			node.updateOptions();
 
+			let option = node.options.get(node.value) ?? node.options.keys().next().value;
+			node.setValue(option, {silent: true});
+
 			if (node.editor) {
 				node.generateDefaultEditor();
 			}


src/primitive.js
```diff
@@ -1316,6 +1316,9 @@ Mavo.observe({id: "primitive"}, function({node, type, attribute, record, element
 		else if (attribute === "mv-options") {
 			node.updateOptions();
 
+			let option = node.options.get(node.value) ?? node.options.keys().next().value;
```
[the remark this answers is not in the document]
LeaVerou: The default value will be taken into account inside the node anyway, so this is not the right place to deal with it, and it would be an undersirable coupling (now every time we touch default values, we also need to look in the code that deals with `mv-options` updates). 

In fact, I wonder if even this line might be duplicating logic that should be elsewhere. Isn't there already logic that sets the value appropriately based on `node.options`? If so, we should be invoking that…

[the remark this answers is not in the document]
DmitrySharabin: @LeaVerou, I could manage to find a place where, from my perspective, this logic should be. Could you please give this PR another shot?

src/primitive.js
```diff
@@ -768,6 +768,12 @@ var _ = Mavo.Primitive = class Primitive extends Mavo.Node {
 				let presentational;
 
 				if (this.options) {
+					if (!this.options.has(value)) {
+						// There is no option corresponding to the value. Try the default one.
+						// If there is no option corresponding to the default value either, use the first option
+						value = this.options.has(this.default)? this.default : this.options.keys().next().value;
```
LeaVerou, line 774: We should not change the value if it doesn't match any of the options, as that's destructive. The user may be experimenting with a different UI, or have forgotten to add all pertinent options. As a design principle, loading data into *any* Mavo and then saving it back should be a non-destructive operation. 

For `<select>` there is a whole code path that creates (and removes when no longer needed) `.mv-volatile` options in the `<select>` to deal with this. `mv-options` should behave similarly.

LeaVerou (changes requested):
See comment

src/primitive.js
LeaVerou: Possibly not for v0.3.0 then?

DmitrySharabin: > Possibly not for v0.3.0 then?

I was going to suggest this. I might have been a bit too optimistic about the issue. 😅

Issue #975, opened 2023-10-03
`Mavo.Script.getVariables(ast)` function to extract expression variables


LeaVerou:
This is part of https://github.com/mavoweb/mavo/issues/967 and depends on #974 

### What would this function do?

This function would extract variables, i.e. non-literal operands, from a Mavo expression and return them as an array of AST nodes (object literals with a `type` attribute).

Requirements:
- AST nodes returned will be either `Identifier` or `MemberExpression` or (more rarely) a `ThisExpression`. However, not all nodes of these types will be part of a variable.
- For `MemberExpression` nodes (e.g. `foo.bar.baz`) we want the topmost `MemberExpression` node.
- Variables cannot be function calls. While in the future we may have some cool functionality where authors can create custom functions (e.g. see #161), don't worry about that right now.

### Step 1: Map generalized AST structures to separate `getVariables()` functions

For each AST structure identified in #974, write a function that maps it to an array of its variables.

For example, for the type of expression in Step 2 of #974, there is only one operand, which is returned by `ast.arguments[0]`,
whereas for an expression of the form `foo(bar + 1)`, `ast.arguments[0]` would be a `BinaryExpression` — there we'd get the variable via `ast.arguments[0].left`.

You will probably notice a lot of patterns here, and will already start consolidating similar AST structures and coming up with a prototype algorithm. This is great!

### Step 2: Create automated tests 

Create a `Mavo.Script.getVariables(ast)` function, with the signature described above, but do not implement it yet.

Add a page in test.mavo.io, pick one expression from each AAST category, and create tests for the function.

### Step 5: Write algorithm to extract variables 

Hopefully, after Step 1, this will be an easy task. 
Use the tests you wrote in Step 2 to verify the algorithm is correct.

LeaVerou:
Implementation sketch, entirely untested, may or may not be correct:

```js
function getVariables(node) {
	let args;
	switch (node.type):
		case "LiteralExpression": 
			return [];
		case "UnaryExpression":
			return getVariables(node.argument);
		case "CallExpression": 
			args = node.arguments; break;
		case "Compound": 
			args = node.body; break;
		case "BinaryExpression":
			args = [node.left, node.right]; break;
		case "ArrayExpression":
			args = node.elements; break;
		case "ConditionalExpression":
			args = [node.test, node.consequent, node.alternate]; break;
		case "ThisExpression":
		case "Identifier":
		case "MemberExpression":
		default:
			return [node];
	}

	if (args) {
[...]

Issue #967, opened 2023-09-02
Migrating away from `with()` so we can move to ESM


LeaVerou:
The primary reason Mavo is not ESM yet is that we use `with()` to evaluate expressions: 

- https://github.com/mavoweb/mavo/blob/master/src/mavoscript.js#L876-L877
- https://github.com/mavoweb/mavo/blob/master/src/mavoscript.js#L899

ESM enforces strict mode, which [does not allow `with()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#removal_of_the_with_statement).

Refactoring our code so that it doesn't use `with()` will not only allow us to move to ESM, but will also make expression evaluation faster (though it's hard to tell how much).

This is not a simple undertaking, but I do think it's possible. We're already rewriting function calls to be called on `$fn` (in the past we used `with()` for that too), we need to do something similar for data as well. Something like this:

1. Find all operands and rewrite them to start from `$data` (e.g. `foo + bar.baz` would become `$data.foo + $data.bar.baz`)
2. Define `$data` right under $fn` to something like: `let $data = data ?? Mavo.Data.stub;
3. Done!

Number 1 is the tricky bit. I've had a lot of false starts trying to write an algorithm to correctly extract all operands from an expression. It seems deceptively simple, but is not if you take all the corner cases into account. Perhaps we need a structured effort:

1. Collect *many* expression examples from a variety of Mavo apps. It should be relatively easy to write a bookmarklet that traverses all `Expression` objects and prints an array in the console. Or even one that appends them to a JSON file, so all we need to do is login (if we're not already logged in), press the button and chill.
2. Pare down these expression examples to those that showcase different AST structures that we need to recognize (e.g. no point in having `foo + bar` and `a * b`, these have the same structure, they're both `BinaryExpression` nodes with a `left` and `right` that are `Identifier` nodes)
3. Convert these into tests
4. At this point it should be easier to come up with a good algorithm and verify that it works well.

Rewriting operands to be chained on `$data` will also make it easier to maybe one day rewrite Mavo to use another reactivity engine under the hood, which will resolve the source of a great chunk of bugs.

PS: We'd need to release a new stable version before any of this work. We [need to fix 3 regressions](https://github.com/orgs/mavoweb/projects/1/views/1?layout=board) for that to happen, none of which seem particularly hard.

karger:
[...]

PR #980, opened 2023-10-13
Create getVariables function; addresses #975

adamjanicki2:
_Draft PR_

## Summary
- Implements a draft version of the function `getVariables`, which lies in `Mavoscript.js`, meaning it can be called by `Mavo.Script.getVariables(ast)`
- Walks an AST and returns a list of all variables used (including duplicates)
- Will help address #967 

## Work still needed
- address `where` clauses: they're hard to address because for an expression like `count(blog.post where cat in category)`, the function would attempt to return `cat` in the list of variables even though it shouldn't
- address js builtins, e.g. `Array.isArray`: you can see in this PR that there's two lines commented out in the `case: "Identifier"`. That's because it doesn't work when a Mavo-evaluated variable gets added to the global scope, and thus `globalThis[variableName]` will be defined (this can be easier discussed offline because it's pretty complicated)

## Testing
- test cases coming soon in a PR; they were scraped from Mavo sites, and the data can be found [here](https://coda.io/d/Mavo-research_d8ZCqXVfZJ4/JSONs_suO7x#_luK1O)

### DIFF
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -554,11 +554,9 @@ var _ = Mavo.Script = {
 			case "LogicalExpression":
 				return [ast.left, ast.right].flatMap(_.getVariables);
 			case "CallExpression":
-				let toExplore = [...ast.arguments];
-				// prevents adding the function name to the list of vars
-				if (ast.callee.type !== "Identifier") {
-					toExplore = [ast.callee].concat(toExplore);
-				}
+				// handles cases like count(...) where the callee is not a variable
+				const toExplore = ast.callee.type === "Identifier"? [] : [ast.callee];
+				toExplore.push(...ast.arguments);
 				return toExplore.flatMap(_.getVariables);
 			case "Compound":
 				return ast.body.flatMap(_.getVariables);
@@ -568,11 +566,14 @@ var _ = Mavo.Script = {
 				return [ast.test, ast.consequent, ast.alternate].flatMap(_.getVariables);
 			case "MemberExpression":
 				const children = _.getVariables(ast.object);
+				const propertyChildren = ast.property.type === "Identifier" ? [] : _.getVariables(ast.property);
 				// If the object is also a member expression, return this one (the top most one)
+				// along with the children of the properties
 				if (children[0] === ast.object) {
-					return [ast];
+					// children.slice(1) adds any variables found in properties of the object or the object's children
+					return [ast].concat(propertyChildren).concat(children.slice(1));
 				}
-				return children;
+				return children.concat(propertyChildren);
 			case "ThisExpression":
 			default:
 				return [ast];
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -534,6 +534,41 @@ var _ = Mavo.Script = {
 		return [a, b];
 	},
 
+	getVariables: function(ast) {
+		const addedVariableNames = new Set();
+
+		function innerGetVariables(ast) {
+			switch (ast.type) {
+				case "Literal":
+					return [];					
+				case "UnaryExpression":
+					return innerGetVariables(ast.argument);
+				case "BinaryExpression":
+					return [ast.left, ast.right].flatMap(innerGetVariables);
+				case "CallExpression":
+					return ast.arguments.flatMap(innerGetVariables);
+				case "Compound":
+					return ast.body.flatMap(innerGetVariables);
+				case "ArrayExpression":
+					return ast.elements.flatMap(innerGetVariables);
+				case "ConditionalExpression":
+					return [ast.test, ast.consequent, ast.alternate].flatMap(innerGetVariables);
+				case "MemberExpression":
+				case "ThisExpression":
+				case "Identifier":
+				default:
+					const name = ast.name;
+					if (!addedVariableNames.has(name)) {
+						addedVariableNames.add(name);
+						return [ast];
+					}
+					return [];
+			}
+		}
+
+		return innerGetVariables(ast);
+	},
+
 	childProperties: [
 		"arguments", "callee", // CallExpression
 		"left", "right", // BinaryExpression, LogicalExpression
--- a/src/mavoscript.js
+++ b/src/mavoscript.js
@@ -533,40 +533,48 @@ var _ = Mavo.Script = {
 
 		return [a, b];
 	},
-
+	/**
+	 * Returns a list of all variables in the AST
+	 */
 	getVariables: function(ast) {
-		const addedVariableNames = new Set();
-
-		function innerGetVariables(ast) {
-			switch (ast.type) {
-				case "Literal":
-					return [];					
-				case "UnaryExpression":
-					return innerGetVariables(ast.argument);
-				case "BinaryExpression":
-					return [ast.left, ast.right].flatMap(innerGetVariables);
-				case "CallExpression":
-					return ast.arguments.flatMap(innerGetVariables);
-				case "Compound":
-					return ast.body.flatMap(innerGetVariables);
-				case "ArrayExpression":
-					return ast.elements.flatMap(innerGetVariables);
-				case "ConditionalExpression":
-					return [ast.test, ast.consequent, ast.alternate].flatMap(innerGetVariables);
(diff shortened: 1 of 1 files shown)


LeaVerou:
On it now, sorry for the delay!

> * address `where` clauses: they're hard to address because for an expression like `count(blog.post where cat in category)`, the function would attempt to return `cat` in the list of variables even though it shouldn't

Why shouldn't it? For that expression, I'd expect it to return nodes for `blog.post`, `cat`, and `category`.
The difficulty with `where` has nothing to do with getting variables, it has to do with rewriting it, since we don't know whether each variable belongs to its left operand (and thus should become `blog.post.cat` or should be evaluated like a regular variable in the scope.

> * address js builtins, e.g. `Array.isArray`: you can see in this PR that there's two lines commented out in the `case: "Identifier"`. That's because it doesn't work when a Mavo-evaluated variable gets added to the global scope, and thus `globalThis[variableName]` will be defined (this can be easier discussed offline because it's pretty complicated)

Not sure I follow: Mavo-evaluated variables (should) never get added to the global scope. I know you said it's complicated but I wonder if I may get it with a few more details.

> * test cases coming soon in a PR; they were scraped from Mavo sites, and the data can be found [here](https://coda.io/d/Mavo-research_d8ZCqXVfZJ4/JSONs_suO7x#_luK1O)

This is great, thanks! But I was also hoping we could have the datasets *in* Coda to make it easier to explore them, e.g. that's why I started this table here: https://coda.io/d/Mavo-research_d8ZCqXVfZJ4/Expressions-dataset_suRYG#_luLla

src/mavoscript.js
```diff
@@ -533,6 +533,49 @@ var _ = Mavo.Script = {
...
+				return toExplore.flatMap(_.getVariables);
+			case "Compound":
+				return ast.body.flatMap(_.getVariables);
+			case "ArrayExpression":
+				return ast.elements.flatMap(_.getVariables);
+			case "ConditionalExpression":
+				return [ast.test, ast.consequent, ast.alternate].flatMap(_.getVariables);
+			case "MemberExpression":
+				const children = _.getVariables(ast.object);
+				if (children[0] === ast.object) {
```
LeaVerou, line 570: Some comments about what this is doing and why would be useful.

src/mavoscript.js
```diff
@@ -533,6 +533,49 @@ var _ = Mavo.Script = {
...
+				if (ast.operator === ':') {
+					return _.getVariables(ast.right);
+				}
+				return [ast.left, ast.right].flatMap(_.getVariables);
+			case "LogicalExpression":
+				return [ast.left, ast.right].flatMap(_.getVariables);
+			case "CallExpression":
+				let toExplore = [...ast.arguments];
+				if (ast.callee.type !== "Identifier") {
+					toExplore = [ast.callee].concat(toExplore);
```
LeaVerou, line 559: Here as well, some comments about what this is trying to detect (e.g. with example expressions)

adamjanicki2:
> On it now, sorry for the delay!
> 
> > * address `where` clauses: they're hard to address because for an expression like `count(blog.post where cat in category)`, the function would attempt to return `cat` in the list of variables even though it shouldn't
> 
> Why shouldn't it? For that expression, I'd expect it to return nodes for `blog.post`, `cat`, and `category`. The difficulty with `where` has nothing to do with getting variables, it has to do with rewriting it, since we don't know whether each variable belongs to its left operand (and thus should become `blog.post.cat` or should be evaluated like a regular variable in the scope.

Think I jumped ahead too quickly, you're right that it doesn't make sense to try and address `where` in here, but later when rewriting expressions 😅 

> 
> > * address js builtins, e.g. `Array.isArray`: you can see in this PR that there's two lines commented out in the `case: "Identifier"`. That's because it doesn't work when a Mavo-evaluated variable gets added to the global scope, and thus `globalThis[variableName]` will be defined (this can be easier discussed offline because it's pretty complicated)
> 
> Not sure I follow: Mavo-evaluated variables (should) never get added to the global scope. I know you said it's complicated but I wonder if I may get it with a few more details.

Here's an example of what I was trying to explain. If you uncomment those lines with the `globalThis` check, it can correctly get variables for something like `Array.isArray(a)`. However, for some reason, when I try getting vars for an expression like `foo + bar`, it only returns `bar` because `globalThis.foo` is defined for some reason (see screenshot below).

Screen Shot 2023-10-20 at 8 46 08 AM

After more digging, it turns out this already exists in the test suite in production. Try going to the [mavoscript tests](https://test.mavo.io/mavoscript.html), inspecting the page, and input `globalThis.foo` into the console. It should return `5`, not `undefined` 😮 You probably have way more insight as to what's going on here, but this is super interesting!


> 
> > * test cases coming soon in a PR; they were scraped from Mavo sites, and the data can be found [here](https://coda.io/d/Mavo-research_d8ZCqXVfZJ4/JSONs_suO7x#_luK1O)
> 
> This is great, thanks! But I was also hoping we could have the datasets _in_ Coda to make it easier to explore them, e.g. that's why I started this table here: https://coda.io/d/Mavo-research_d8ZCqXVfZJ4/Expressions-dataset_suRYG#_luLla

Yeah, I haven't had time to copy and paste all the expressions out into that table yet, but should be able to get around to it today or tomorrow 👍

src/mavoscript.js
adamjanicki2: Sounds good, will add better, descriptive comments here (and same thing for below)!

LeaVerou:
> After more digging, it turns out this already exists in the test suite in production. Try going to the [mavoscript tests](https://test.mavo.io/mavoscript.html), inspecting the page, and input `globalThis.foo` into the console. It should return `5`, not `undefined` 😮 You probably have way more insight as to what's going on here, but this is super interesting!

That is incredibly bizarre, since that page is not evaluating expressions at all, only parsing them and rewriting them 🤔🤔🤔

Can you narrow down when this happens? E.g.:
- Can you reproduce it outside of the testsuite? E.g. in a codepen? You can use play.mavo.io as a starting point. 
- Does it still happen if you comment out all the tests? If not, is there a specific test that makes it happen?

You should uncomment this part of your code, and we should instead track down why this happens.
We can't merge this PR until we release v0.3.0 anyway (though that's [almost done](https://github.com/orgs/mavoweb/projects/1?query=is%3Aopen+sort%3Aupdated-desc)).

> Yeah, I haven't had time to copy and paste all the expressions out into that table yet, but should be able to get around to it today or tomorrow 👍

It would be nice to also have a simplified AST for them. Note that importing them into Coda can be very fast: as long as you generate rows separated by line breaks and values separated by tabs you can just paste data in a table and it works. Importing a CSV works too (comma-separated values, line break separated rows).

adamjanicki2:
👍 I will find where this issue is stemming from + do everything else you mentioned in the above comment hopefully by end of today

adamjanicki2:
Found the culprit: https://github.com/mavoweb/test/blob/master/mavoscript.html#L230

So looks like the approach should work!

src/mavoscript.js
```diff
@@ -533,6 +533,51 @@ var _ = Mavo.Script = {
...
+				}
+				return [ast.left, ast.right].flatMap(_.getVariables);
+			case "LogicalExpression":
+				return [ast.left, ast.right].flatMap(_.getVariables);
+			case "CallExpression":
+				let toExplore = [...ast.arguments];
+				// prevents adding the function name to the list of vars
+				if (ast.callee.type !== "Identifier") {
+					toExplore = [ast.callee].concat(toExplore);
+				}
```
LeaVerou, line 561: Nit, but I found this code a little hard to read.

If you want to add an element to the beginning of an array, you can use [`Array#unshift()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift)

But it may be better to add things in the right order to begin with, rather than having to change it later, e.g. like this:
```suggestion
				// For plain function calls (e.g. foo()), the callee is not a variable
				let toExplore = ast.callee.type === "Identifier" ? [] : [ast.callee];
				toExplore.push(...ast.arguments);
```

PR #981, opened 2023-10-15
Make the disabled attribute boolean on all supported elements

DmitrySharabin:
For details, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled#browser_compatibility


LeaVerou (approved):

LeaVerou:
That probably didn't need a review at all but approved anyway :)

PR #982, opened 2023-10-19
Functionality improved of ui.popup.js

Akshat111111:
Simplified DOM manipulation using element.appendChild() and element.remove(),Added Arrow function,Removed unnecessary variables and added ES6 classes.

### DIFF
--- a/src/ui.popup.js
+++ b/src/ui.popup.js
@@ -1,140 +1,113 @@
-(function($, $$) {
+class Popup {
+  constructor(primitive) {
+    this.primitive = primitive;
+    this.shown = false;
+    this.editor = primitive.editor;
+    this.element = document.createElement("div");
+    this.element.classList.add("mv-popup");
+    this.element.hidden = true;
 
-var _ = Mavo.UI.Popup = $.Class({
-	constructor: function(primitive) {
-		this.primitive = primitive;
+    this.position = () => {
+      const bounds = this.primitive.element.getBoundingClientRect();
+      const x = bounds.left;
+      let y = bounds.bottom;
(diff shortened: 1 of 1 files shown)


LeaVerou (changes requested):
Hi @Akshat111111 ! Thank you for working on contributing to Mavo. Some comments below.

1. In general we don’t make such extensive changes just for refactoring purposes, as they run the risk of introducing bugs on areas of code we may have not touched for a while, and pollute the commit log with non substantive changes. Instead, we tend to refactor as we go, updating pieces of code when we actually work on them.
2. Your PR appears to have changed indentation from tabs to spaces, likely due to editor settings. It’s good practice to review your diff before committing, it helps catch a lot of these issues early.
3. It appears you have removed some comments? Comments are there to explain what the code does to new contributors (like you!), `TODO` comments are there to remind us about things that need doing, `FIXME` comments about caveats etc. Is there a reason you felt they may not be useful?
4. Curious why you feel the (older) `appendChild()` is somehow better than `append()`?
5. I’m all for removing some Bliss calls now that the DOM is in better shape, but it needs to be done with care to preserve semantics and functionality. E.g. `$.remove(el)` is not equivalent to `el.remove()`: the latter will throw if `el` is null/undefined. It needs to be rewritten as `el?.remove()`.

Given (1) above, I’m going to go ahead and close this, but please don’t be discouraged from future contributions; there’s always a bit of a learning curve when contributing to a new project 😊

Akshat111111:
Yeah, Thankyou so much for your descriptive review,I understand it very closely and really learnt a lot from it.

PR #985, opened 2023-10-24
URL params should have precedence over storage attributes, closes #984

### LINKED ISSUE #984: A backend specified via the URL should have priority over backend metadata passed via storage attributes
As we state [in the docs](https://mavo.io/docs/storage#url-params), the user can change the storage location of any Mavo app via the URL.

On the other hand, the author can configure the backend they use via [storage attributes](https://mavo.io/docs/storage#storage-attributes). And **those attributes have priority** over metadata Mavo can get by parsing the storage URL.

In most cases, that works perfectly fine. However, those storage attributes **also have priority over the URL** the user might pass to override the default storage. Consequently, the user might be unable to store their data in the file specified via the URL. So, it might break the brilliant idea of re-using apps.

For example, suppose the page with our app has the URL `https://mavo.app`, and the app is defined like so:

```html
<div mv-app mv-storage="https://github.com/mavoweb/mavo.io/docs/docs.json" mv-storage-branch="master">
	<!-- Our awesome app -->
</div>
```

This app has the following `primaryBackend` (mind `branch` and `source`):

image

If we then want to work with data stored in another branch on the same repo, we can pass the file URL via the `storage` URL param like so: `https://mavo.app?storage=https://github.com/mavoweb/mavo.io/blob/road-to-0.3.0/docs/docs.json` (mind that the branch name—`road-to-0.3.0`—is in the URL).

However, the `branch` property of the `primaryBackend` didn't change (since the storage attributes have priority):

image

That means that Mavo will store data in the file with the following URL: `https://github.com/mavoweb/mavo.io/blob/master/docs/docs.json` which is (technically) equivalent to `https://github.com/mavoweb/mavo.io/docs/docs.json` (the same URL that we provided authoring the app). As a result, **we couldn't override the app's storage**.

DmitrySharabin:


src/mavo.js
```diff
@@ -426,16 +426,30 @@ let _ = self.Mavo = $.Class(class Mavo {
...
 			// This app is the first one in the page, so we can override its backend
 			// via URL params such as ?storage=...
 			backend = _.Functions.url(role);
+
+			if (backend) {
+				ignoreRoleAttributes = true;
+			}
 		}
 
 		if (!backend) {
```
joyously, line 444: Was this if statement working before, with `backend` uninitialized?

LeaVerou: Presumably it would always be true unless `this.index == 1`

DmitrySharabin: AFAIK this works precisely as Lea described.

src/mavo.js
```diff
@@ -426,16 +426,30 @@ let _ = self.Mavo = $.Class(class Mavo {
...
-			backend = _.Functions.url(`${this.id}-${role}`)
-			          || _.Functions.url(`${this.id.toLowerCase()}-${role}`)
-			          || this.element.getAttribute(attribute) || null;
+			backend = _.Functions.url(`${this.id}-${role}`) || _.Functions.url(`${this.id.toLowerCase()}-${role}`);
```
LeaVerou, line 445: Unrelated but this should be part of the `url()` function, case insensitivity would be useful more broadly. In fact, it should probably be the default, with some escape hatch when people want case sensitive matching.

LeaVerou (changes requested):
This seems a little roundabout. Correct me if I'm mistaken but it looks like you're setting a flag, then reading the attributes normally, but ignoring them if the flag is set. Why not just ...not read them in the first place if the backend is provided via the URL?

DmitrySharabin:
> This seems a little roundabout. Correct me if I'm mistaken but it looks like you're setting a flag, then reading the attributes normally, but ignoring them if the flag is set. Why not just ...not read them in the first place if the backend is provided via the URL?

You are absolutely right. I just wanted to change the existing code as little as I could. But yeah, it's much better not to read attributes at all if the backend is provided via the URL. Will fix it. And will use the new `url()` function a bit later.

karger:
Coming to this late, but *why* should url params have precedence?   If there are conflicts between url params and storage attributes, it's easy to remove the conflicting attributes, but harder to manipulate the url to eliminate the conflicting params.   So isn't it easier on the author to let them override params they don't like with attributes, instead of telling them that they can override attributes they don't like with params?

DmitrySharabin:
> Coming to this late, but _why_ should url params have precedence? If there are conflicts between url params and storage attributes, it's easy to remove the conflicting attributes, but harder to manipulate the url to eliminate the conflicting params. So isn't it easier on the author to let them override params they don't like with attributes, instead of telling them that they can override attributes they don't like with params?

I was trying to explain it here: #984

LeaVerou:
@karger it looks like you deleted your comment, so this may not be needed, but in case: You don't necessarily have control of a page for which you're trying to override storage via URL params. The whole point of this feature was to allow repurposing of apps that you did not create. If they don't take precedence, the feature is useless.

DmitrySharabin:
> This is better but what I had in mind didn't need a flag at all: read the URL params, later down check if you already have a backend before you even look at attributes. (Yes, this means that line 451 will have to be moved)

@LeaVerou It look like I could find a way to eliminate the `ignoreRoleAttributes` flag. Would you mind taking a look?

src/mavo.js
```diff
@@ -433,25 +434,29 @@ let _ = self.Mavo = $.Class(class Mavo {
...
-			backend = _.Functions.url(`${this.id}-${role}`)
-			          || _.Functions.url(`${this.id.toLowerCase()}-${role}`)
-			          || this.element.getAttribute(attribute) || null;
+			backend = _.Functions.url(`${this.id}-${role}`);
 		}
 
-		if (backend) {
-			backend = backend.trim();
+		if (!backend) {
+			backend = this.element.getAttribute(attribute) || null;
```
LeaVerou, line 441: I'd move `const attribute = "mv-" + role;` in here if it's not used anywhere outside this.

LeaVerou (approved):
Yes, this is roughly what I had in mind! Well done!

src/mavo.js
DmitrySharabin: Right. Done. Thanks!

PR #988, opened 2023-10-25
Better url()

LeaVerou:
See https://github.com/mavoweb/mavo/pull/985/files#r1371833984

- Add options (type, case_sensitive, multiple). Note the design with the multiple options objects allows calling it without `group()`, like `url(foo, bar, case_sensitive: true, multiple: true, type: 'path')`
- Made matching case insensitive by default
- Do not strip non-ASCII chars, just escape those that you need to escape (i18n yo)
- Refactor

Has not been tested, @DmitrySharabin you need to run it (and check the tests) before merging! Feel free to make any edits you want.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -176,7 +176,7 @@ let _ = Mavo.Functions = {
 
 
 				if (index > -1) {
-					let ret = path[index + 1];
+					let ret = path[(index + 1) % path.length]; // Why “% path.length”? To handle cases like “/foo” where foo is equal to "".
 
 					if (ret) {
 						ret = decodeURIComponent(ret);
--- a/src/functions.js
+++ b/src/functions.js
@@ -176,7 +176,7 @@ let _ = Mavo.Functions = {
 
 
 				if (index > -1) {
-					let ret = path[(index + 1) % path.length]; // Why “% path.length”? To handle cases like “/foo” where foo is equal to "".
+					let ret = path[index + 1] ?? "";
 
 					if (ret) {
 						ret = decodeURIComponent(ret);
--- a/src/functions.js
+++ b/src/functions.js
@@ -129,8 +129,8 @@ let _ = Mavo.Functions = {
 	/**
 	 * Get the page URL (if called with no params), or a parameter from it.
 	 * @param {string} id Parameter name
-	 * @param {string | URL} [url] URL to get parameter from. Defaults to current page URL
 	 * @param {object} [...options] One or more options objects
+	 * @param {string | URL} [options.url] URL to get parameter from. Defaults to current page URL
 	 * @param { "query" | "path" } [options.type] Type of parameter to get.
 	 * 		"query" for query string type URLs (?foo=value)
 	 * 		"path" for path type URLs (/foo/value)
@@ -139,7 +139,7 @@ let _ = Mavo.Functions = {
 	 * @param {boolean} [options.multiple=false] Whether to return multiple values if there are multiple parameters with the same name
 	 * @returns {string | null}
 	 */
-	url: (id, url = location, ...options) => {
+	url: (id, ...options) => {
 		if (id === undefined) {
 			// url() with no arguments is just an alias for location.href
 			return location.href;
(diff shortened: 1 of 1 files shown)


src/functions.js
```diff
@@ -126,23 +126,71 @@ let _ = Mavo.Functions = {
...
+	 * @param {object} [...options] One or more options objects
+	 * @param { "query" | "path" } [options.type] Type of parameter to get.
+	 * 		"query" for query string type URLs (?foo=value)
+	 * 		"path" for path type URLs (/foo/value)
+	 * 		If not specified, it will search for both, with query string parameters taking precedence
+	 * @param {boolean} [options.case_sensitive=false] Whether to do a case sensitive search
+	 * @param {boolean} [options.multiple=false] Whether to return multiple values if there are multiple parameters with the same name
+	 * @returns {string | null}
+	 */
+	url: (id, url = location, ...options) => {
```
DmitrySharabin, line 142: Why do we need the `url` parameter here? I [searched through our codebase](https://github.com/search?q=repo:mavoweb/mavo%20.url(&type=code) and found no places where we use it. Moreover, [we didn't document](https://mavo.io/docs/functions/#url) that `url()` has an optional second parameter. What am I missing?

If we leave it as it is, we should provide a value for it every time we need to perform, e.g., a case-insensitive search, like so: `url("foo", location, case_insensitive: true)`.

DmitrySharabin: I think I get why we might need it. But what if we make `url` one of the options, too, so we don't have to provide it all the time? Then we can analyze any URLs (not only the one of the current page) with the `url()` function.

LeaVerou: Fair enough, it's totally possible to want to specify options, and you shouldn't also need to specify the URL. Can you make the change?

DmitrySharabin: Already did; just waited for your answer to push. :)

LeaVerou: Looks good!

adamjanicki2: Nice!

karger:
How does providing all these parameters to the url() function compare to just having it return an object whose properties correspond to the parts of the url, that can be extracted with dot notation?

LeaVerou:
> How does providing all these parameters to the url() function compare to just having it return an object whose properties correspond to the parts of the url, that can be extracted with dot notation?

What kind of object structure are you picturing? And what does it do in the common case where it's used as a primitive?

src/functions.js
```diff
@@ -126,23 +126,71 @@ let _ = Mavo.Functions = {
...
-		if (ret === null || !id) {
-			return null;
+			if (type === "path" || !type) {
+				let path = url.pathname.split("/");
+				// FIXME support multiple here?
+				let index = case_sensitive ? path.indexOf(id) : path.findIndex(part => part.toLowerCase() === id.toLowerCase());
+
+
+				if (index > -1) {
+					let ret = path[(index + 1) % path.length]; // Why “% path.length”? To handle cases like “/foo” where foo is equal to "".
```
LeaVerou, line 179:
```suggestion
					let ret = path[index + 1] ?? "";
```

karger:
> > How does providing all these parameters to the url() function compare to just having it return an object whose properties correspond to the parts of the url, that can be extracted with dot notation?
> 
> What kind of object structure are you picturing? And what does it do in the common case where it's used as a primitive?

Basically the DOM Location object.   Which could include the .href to provide the full url.   Or perhaps it just gives you the url string when used in a string context?

LeaVerou:
> > > How does providing all these parameters to the url() function compare to just having it return an object whose properties correspond to the parts of the url, that can be extracted with dot notation?
> > 
> > 
> > What kind of object structure are you picturing? And what does it do in the common case where it's used as a primitive?
> 
> Basically the DOM Location object. Which could include the .href to provide the full url. Or perhaps it just gives you the url string when used in a string context?

This does not return a location, so not sure how that relates. Why would it provide the full url when its purpose is to read parameters from the URL? Also, objects tend to be hard for novices.

PR #991, opened 2023-11-02
Ignore un-supported events that backend services might send

DmitrySharabin:
Working on [the new Yandex Disk backend](https://github.com/DmitrySharabin/mavo-yandex), I noticed that it sends an extra message event to the `authPopup` window while authenticating the user. Yandex Metrics sends this message. The event doesn't contain the `backend` field in the payload, neither does it contain `accessToken` and breaks the login flow since we [`reject(Error("Authentication error"))`](https://github.com/mavoweb/mavo/blob/cf64cf58f66fa7553a3c9829153fc5724416bafa/src/backend.js#L201-L203) in such cases.

Let's ignore all `message` events that don't serve the authentication purpose.

### DIFF
--- a/src/backend.js
+++ b/src/backend.js
@@ -193,12 +193,7 @@ var _ = Mavo.Backend = class Backend extends EventTarget {
 					}
 
 					addEventListener("message", evt => {
-						if (evt.source === this.authPopup) {
-							if (!evt.data.backend) {
-								// Ignore un-supported events that backend services might send, e.g., messages from Yandex metrics.
-								return;
-							}
-
+						if (evt.source === this.authPopup && evt.data.backend) {
 							if (evt.data.backend == this.id) {
 								this.accessToken = localStorage[`mavo:${id}token`] = evt.data.token;
 							}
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
It looks like Madata also suffers from this issue.

LeaVerou (approved):
+1, this didn't even need a review IMO

LeaVerou:
> It looks like Madata also suffers from this issue.

I’m not surprised.

In fact, we're so close to being able to use ESM in Mavo, that I'm wondering if there's much point in creating new Mavo backends at all.

DmitrySharabin:
> In fact, we're so close to being able to use ESM in Mavo that I'm wondering if there's much point in creating new Mavo backends at all.

People from the Russian Mavo community asked me to build it. And since I would love them to continue using Mavo, I agreed. ☺️
If it works in Mavo well (as a starting point), there is not much effort to port it to Madata later. From my perspective, it's a win-win.

DmitrySharabin:
> I’m not surprised.

I can fix it there too if you don't mind.

LeaVerou:
> > I’m not surprised.
> 
> I can fix it there too if you don't mind.

Please do :)

PR #995, opened 2023-11-16
Disable auto-saving before rendering data only

### LINKED ISSUE #994: Live storage attributes break the autosave feature
This is why @karger's app doesn't work as it should. The reason for this issue is precisely the same as described in [the issue we already closed](https://github.com/mavoweb/mavo/issues/910).

Reduced testcase: https://codepen.io/dmitrysharabin/pen/yLZPeEa

DmitrySharabin:
It seems to me that the only part of the `load()` method code that needs to be executed with the auto-saving disabled is `render()`. Previously, we dropped the `autoSave` flag a bit too early, which might lead to undesired consequences if some of the asynchronous events occur. This patch fixes it (as our tests and my experiments show).

- Fixes #994
- The previous patch for #910 is not needed anymore since we solved the general issue


LeaVerou (approved):
Nice, thanks!

PR #1001, opened 2023-11-27
delete with and move to use vastly

adamjanicki2:
_Work in progress, not ready for review yet_

Addresses #967


src/mavoscript.js
```diff
@@ -538,323 +520,11 @@ var _ = Mavo.Script = {
...
-	childProperties: [
-		"arguments", "callee", // CallExpression
-		"left", "right", // BinaryExpression, LogicalExpression
-		"argument", // UnaryExpression
-		"elements", // ArrayExpression
-		"test", "consequent", "alternate", // ConditionalExpression
-		"object",  "property", // MemberExpression
-		"body"
-	],
+	walk: Vastly.walk,
```
LeaVerou, line 523: @DmitrySharabin what do you think? Should we continue to define these or just use vastly methods internally and print out a deprecation warning for these, directing plugin authors to use vastly too (using your [`Mavo.Functions.deprecatedFunction()`](https://github.com/mavoweb/mavo/blob/main/src/functions.js#L70))?

src/mavoscript.js
```diff
@@ -538,323 +520,11 @@ var _ = Mavo.Script = {
...
-		else if (ret !== undefined) {
-			return ret;
-		}
-
-		if (!node.type || !_.serializers[node.type]) {
-			throw new TypeError("Cannot understand this expression at all 😔");
-		}
-
-		return _.serializers[node.type](node, parent);
-	},
```
LeaVerou, line 857: Oooh that must have been pleasant 😁

However, we still need the transformations (vastly has an export of an empty transformations object so the developer can modify them). Keep in mind vastly is written for parsing/serializing regular JS, so all the special Mavo stuff will still live here.

LeaVerou: Try `Object.assign(Vastly.transformations, transformations)` but do check where rollup puts that export, it may be `Vastly.serialize.transformations` or something entirely different.

adamjanicki2: Followup question: where should this go such that it only runs once? Somewhere around [here](https://github.com/mavoweb/mavo/blob/main/src/mavo.js#L8)?

adamjanicki2: Update: `transformations` and other non-default exports are not accessible outside Vastly. I assume it's because the index file only exports the defaults for each file: https://github.com/mavoweb/vastly/blob/main/src/index.js#L2C1-L9C55

LeaVerou: > Update: `transformations` and other non-default exports are not accessible outside Vastly. I assume it's because the index file only exports the defaults for each file: [mavoweb/vastly@`main`/src/index.js#L2C1-L9C55](https://github.com/mavoweb/vastly/blob/main/src/index.js?rgh-link-date=2023-11-28T19%3A32%3A56Z#L2C1-L9C55)

That's something to fix on the vastly side :) Please file an issue first, so you have something to close.

One way is to add it as a property of `serialize`, another to add it as a separate named export. Which solution would you pick and why? :)

adamjanicki2: Yeah I agree, I was asking where _in mavo_ should we put the line Object.assign(Vastly..., transformations). I suggested [here](https://github.com/mavoweb/mavo/blob/main/src/mavo.js#L8) since there's a very similar line right below to set something inside stretchy

adamjanicki2: > That's something to fix on the vastly side :) Please file an issue first, so you have something to close.

Will do

> One way is to add it as a property of `serialize`, another to add it as a separate named export. Which solution would you pick and why? :)

I'm leaning toward adding it as a prop of `serialize`, because adding additional lines like `export {transformations} from "./serialize.js"` doesn't seem too good to me. When you say add as a property of serialize, how would I go about doing this?

LeaVerou: Ok, let me elaborate a bit, as I think my earlier message wasn't super clear.

Option 1 is that `index-fn` also does `export { transformations, serializers }`. Then they become top-level exports, same as the functions, and they'd be available as e.g. `Vastly.transformations`. 

Option 2 is that we add them as properties of the `serialize` function, since functions are objects in JS:

```
serialize.transformations = transformations;
```

Then they are not exported per se, but can be accessed via `serialize`.

What do you think are the pros & cons of each of these approaches?

adamjanicki2: That's cool! To me, it seems like the pros of option1 is we don't waste space inside of serialize "object", and the pros of option2 is we don't have to worry about exporting a separate thing from index, and `transformations` is tied to `serialize` since that's where it belongs

src/mavoscript.js
DmitrySharabin: I checked all Mavo plugins. Not only ours but also every plugin in the plugin registry. I couldn't find any that uses these `Mavo.Script` methods: `walk`, `closest`, and `serialize`. However, even though we don't use it in the plugins from the registry, we can't guarantee no (private) plugin does. So, I would follow the route with a deprecation message (until the next release minimum). At least we can lower the chance of breaking some author code and level up the awareness of vASTly, which is nice from my perspective. 😊

PR #1003, opened 2023-12-05
migrate to using vastly in mavoscript

adamjanicki2:


src/mavoscript.js
```diff
@@ -628,7 +604,7 @@ var _ = Mavo.Script = {
 			return `${nameSerialized}(${argsSerialized.join(", ")})`;
 		},
 		"ConditionalExpression": node => `${_.serialize(node.test, node)}? ${_.serialize(node.consequent, node)} : ${_.serialize(node.alternate, node)}`,
-		"MemberExpression": (node, parent) => {
```
adamjanicki2, line 631: I deleted this because it was an unused variable

src/mavoscript.js
```diff
@@ -819,41 +795,13 @@ var _ = Mavo.Script = {
...
-
 		if (parent) {
-			node.parent = parent;
-		}
-
-		var ret = _.transformations[node.type]?.(node, parent);
-
-		if (typeof ret == "object" && ret?.type) {
-			node = ret;
+			Vastly.parents.set(node, parent);
```
adamjanicki2, line 802: I originally wanted to write this function as:
```js
serialize: (node) => {
    Vastly.parents.setAll(node);
    return Vastly.serialize(node);
},
```
However, due to `serializers` needing to pass a parent argument back into `serialize`, we have to do
```js
serialize: (node, parent) => {
    if (parent) {
        Vastly.parents.set(node, parent);
    }
    return Vastly.serialize(node);
},
```

LeaVerou: But is that parent argument used? It seems to be primarily used to set the parent, which we already do.

adamjanicki2: That's what I thought too, but it fails the majority of the serialize tests when implented like that, I'm still working on trying to figure out why that's the case

adamjanicki2: I will update once I figure it out

LeaVerou: Which tests does it fail?

adamjanicki2: Screen Shot 2023-12-05 at 5 27 48 PM

It's all related to member expressions

LeaVerou: Hm, so it seems all related to it rewriting Mavo functions to use `get()` as well, so e.g. instead of `$fn.eq()` you get `$fn.get($fn, "eq")`, which happens [here](https://github.com/mavoweb/mavo/blob/main/src/mavoscript.js#L641). So you need to verify that what you're doing is actually setting parents, cause them not being set properly would be my top guess (or maybe they're not properly set after transformations).

adamjanicki2:
Going to come back to this now

src/mavoscript.js
adamjanicki2: @LeaVerou  Coming back to this bug, it seems the root cause is that serializers mutate the node, meaning the parents originally set by vastly are no longer correct, an example is [here](https://github.com/mavoweb/mavo/blob/main/src/mavoscript.js#L617). 

I think having
```js
if (parent) {
        Vastly.parents.set(node, parent);
}
```
isn't the worst solution, but I'll try to think of a way around it

LeaVerou: Note that what a lot of this logic does is trying to find the topmost variable, which you already have a function for!
Then, it's rewriting to prepend with `$fn`, which seems useful enough in itself that might be worth including in vastly (in a more generic form). E.g. we'll definitely need this for rewriting data variables to branch off `$data` as well, and for rewriting `where` calls (to branch off a function call with the possible alternatives). If this is handled by vastly, parents can be adjusted by vastly and calling code doesn't have to care about maintaining parent pointers.

adamjanicki2: I'll create an issue for it over in vastly

adamjanicki2: https://github.com/mavoweb/vastly/issues/32

Issue #996, opened 2023-11-17
Linked dynamic <select> is broken if the app starts in edit mode


DmitrySharabin:
Reduces testcase: https://codepen.io/dmitrysharabin/pen/ExroXGa
First bad commit: https://github.com/mavoweb/mavo/commit/9f7876e32ff03b33b4f6cf2295c9d14aaacd149b

DmitrySharabin:
It looks like we have this issue not only with linked `<select>`s but also with [implicit booleans](https://codepen.io/dmitrysharabin/pen/WNmKrML).

**Expected behavior**

<img width="111" alt="image" src="https://github.com/mavoweb/mavo/assets/9166277/267ed180-fb7e-43a3-9b16-b6a1259d18e5">

**Current behavior**

<img width="213" alt="image" src="https://github.com/mavoweb/mavo/assets/9166277/acd6fec7-c2fd-4c50-81d4-0662a0f3eaa1">

It feels like we should give Mavo some time to settle down all the things with properties (or some sort of “dependencies”) before switching to the edit mode.

PR #1006, opened 2024-01-03
Preserve the original editor observer

DmitrySharabin:
Fixes regression #996.

The main issue is that we lose the original editor observer when we force the original editor update, so the changes are not picked up when the original editor is updated via an expression.

### DIFF
--- a/src/primitive.js
+++ b/src/primitive.js
@@ -426,7 +426,14 @@ var _ = Mavo.Primitive = class Primitive extends Mavo.Node {
 					}
 
 					for (let primitive of nodes) {
-						primitive.originalEditorUpdated({force: true});
+						if (primitive.defaultSource == "editor") {
+							primitive.default = this.originalEditor.value;
+						}
+
+						if (primitive.editor) {
+							primitive.editor = this.originalEditor.cloneNode(true);
+						}
+
 						primitive.setValue(primitive.value, {force: true, silent: true});
 					}
 				});


src/primitive.js
```diff
@@ -426,7 +426,14 @@ var _ = Mavo.Primitive = class Primitive extends Mavo.Node {
 					}
 
 					for (let primitive of nodes) {
-						primitive.originalEditorUpdated({force: true});
```
LeaVerou, line 429: Was this line removed accidentally or we genuinely don't need it anymore?

src/primitive.js
```diff
@@ -441,7 +448,7 @@ var _ = Mavo.Primitive = class Primitive extends Mavo.Node {
 
 		let editor = this.editor ?? this.originalEditor;
 
-		if (editor?.matches("select:not(.mv-options-select")) {
+		if (editor?.matches("select:not(.mv-options-select)")) {
```
LeaVerou, line 451: Whoa, how did this go undetected all this time?!

src/primitive.js
DmitrySharabin: My investigation shows that this line is the one that was added in [the first bad commit](https://github.com/mavoweb/mavo/commit/9f7876e32ff03b33b4f6cf2295c9d14aaacd149b) and caused the regression. The only thing I did was revert the previous version of this code block.

I think I understand why this line was present—we get the same result as when we have the two added lines. However, to add to this, we destroy the original editor observer and replace it with the observer of the clone of the original editor. It works fine if the element is not in the edit mode initially. But when it is, we clone the original editor before Mavo can update expressions and handle the original editor mutations.

src/primitive.js
DmitrySharabin: We have it not so long as you might imagine—not longer than we have `mv-options`. 😉 We couldn't detect it thanks to the browser forgiveness, I believe. I checked in the console—both variants return the same result.

src/primitive.js
DmitrySharabin: It looks like we need this line since, without it, we break another test with dynamic selects. I should continue working on a proper fix.

PR #1007, opened 2024-01-04
Correctly handle <option selected>

### LINKED ISSUE #702: If the data source for dynamic select is asynchronous, the select takes its value from the last option
Even though the browser renders the corresponding `select` element correctly.

Reduced testcase: https://codepen.io/dmitrysharabin/pen/ExNPKqw?editors=1010

This is the reason why https://mavo.io/get/ still doesn't behave the way we expect (relative to https://github.com/mavoweb/mavo/issues/698).

DmitrySharabin:
Fixes #702 (it also might fix some other issue concerning dynamic `<select>`s; I will check).

As [we know from MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#selected), the `selected` boolean attribute indicates that the option is _initially_ selected ([when the page first loads](https://arc.net/l/quote/eehrfvfq)). It means that if we try to change the selected option programmatically (via an expression) after the page is loaded, we'll fail. This PR should solve the issue by telling the corresponding `<select>` that one of its `<option>`s became selected via the `selected` attribute. This will allow Mavo to pick up updates on the corresponding properties.

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -253,6 +253,17 @@ _.register({
 		default: true
 	},
 
+	"option[selected]": {
+		selector: "select option[selected]",
+		attribute: "selected",
+		datatype: "boolean",
+		setValue: (element, value) => {
+			element.selected = value;
+			element.closest("select")?.dispatchEvent(new Event("change"));
+		},
+		getValue: element => element.selected
+	},
+
 	"textarea": {
 		extend: "formControl",
 		selector: "textarea",


LeaVerou:
Hey, thanks for working on this problem!

Not sure about this. This is basically telling Mavo that if `selected` is present on an `<option>`, then the value of that option (by default) is a boolean, not `option.value`. I worry this could cause all sorts of unexpected behavior. At the very least we should test this very very extensively before merging.

DmitrySharabin:
> Hey, thanks for working on this problem!
> 
> Not sure about this. This is basically telling Mavo that if `selected` is present on an `<option>`, then the value of that option (by default) is a boolean, not `option.value`. I worry this could cause all sorts of unexpected behavior. At the very least we should test this very very extensively before merging.

Oh. Right. I get what you mean. Then, we should probably find a better way to monitor the `select` attribute mutation. 😊

Thank you!

PR #1020, opened 2024-02-27
First pass on moving from ResizeObserver to Container Queries

### LINKED ISSUE #1018: [Refactor] Use container queries for the Mavo toolbar
There’s a bit of code in `src/ui.bar.js` to adapt the Mavo toolbar to different sizing conditions. We should simply use container queries for it, since these days they’re [supported everywhere](https://caniuse.com/css-container-queries) (Baseline 2023!).

This may be a good issue for someone who has been looking for an excuse to try container queries 😄

DmitrySharabin:
Closes #1018.

https://github.com/mavoweb/mavo/assets/9166277/7700dacb-9c37-441c-9ccc-88d2757f8958

### DIFF
--- a/src-css/_bar.scss
+++ b/src-css/_bar.scss
@@ -11,6 +11,8 @@
 	line-height: 1.5;
 	position: sticky;
 
+	container: mv-bar / inline-size;
+
 	@supports (position: sticky) {
 		z-index: 2;
 		top: 5px;
@@ -175,20 +177,25 @@
 	}
 
 	// Displaying mv-bar in small spaces
-	&.mv-compact {
-		.mv-status > span {
-			display: none;
+	@container mv-bar (width <= 40em) {
+		.mv-status {
+			// Don't wrap the username on multiple lines
+			white-space: nowrap;
+
+			& > span {
+				display: none;
+			}
 		}
 	}
 
-	&.mv-tiny {
+	@container mv-bar (width <= 25em) {
 		& > button,
 		& > .mv-button {
 			// Hide text
 			width: 1em;
 			position: relative;
 			overflow: hidden;
-			text-indent: -999em;
+			text-indent: calc(-infinity * 1em);
 
 			&::before {
 				position: absolute;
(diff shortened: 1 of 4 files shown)


src/ui.bar.js
```diff
@@ -100,57 +92,9 @@ let _ = Mavo.UI.Bar = class Bar {
...
-			this.resize();
-
-			if (self.ResizeObserver) {
-				this.resizeObserver = Mavo.observeResize(this.element, entries => {
-					this.resize();
-				});
-			}
-		}
-
 		Mavo.observers.resume();
```
DmitrySharabin, line 95: Do we still need it here?

src/ui.bar.js
DmitrySharabin, line 1: I wonder, do we still need these two lines? Do we still have observers that need to be paused and resumed here? 🤔

(line 12)
```js
Mavo.observers.pause();
```
(line 95)
```js
Mavo.observers.resume();
```

src-css/_bar.scss
```diff
@@ -175,20 +177,25 @@
 	}
 
 	// Displaying mv-bar in small spaces
-	&.mv-compact {
-		.mv-status > span {
-			display: none;
+	@container mv-bar (width <= 40em) {
```
DmitrySharabin, line 180: `40em` and `25em` are magic numbers obtained from experiments that “feel OK” in general cases. However, it's not the silver bullet for all the cases. If the Mavo bar has many buttons, these numbers need to be adjusted.

LeaVerou: One thing at a time. It's ok to keep the magic numbers for now. Eventually we want to move to a component that handles its own layout.

src/ui.bar.js
LeaVerou: I have no idea. Let's leave it for now, can't hurt.

LeaVerou (approved):
LGTM. That felt good, right? 😁

DmitrySharabin:
> LGTM. That felt good, right? 😁

Yeah. Fells like magic. I’m glad I could work on it. We live in the future now. 😄

PR #1025, opened 2024-07-01
Fix ReferenceError: closestItem is not defined

C-Ezra-M:
Occurred to me while trying to use list-separator. This is a temporary solution until this section of code can be properly DRYfied.


DmitrySharabin (approved):
Oh, wow. How on earth we didn't notice this earlier? Thank you so much for finding and fixing this.

A bit of context: this code handles the case when one tries to paste multiple items in a list item. I'm not sure how long this feature has been broken. 🤷‍♂️

C-Ezra-M:
I believe it's ready to merge. I know there was a force push, but it didn't affect `primitive.js`

C-Ezra-M:
I'd like to ask if this will ever be merged. It appears that Mavo has not been maintained since 2024.

DmitrySharabin:
Hey there,
This PR seems to have changes from other PRs. Could you pull the changes from the `main` branch and rebase?

Issue #1036, opened 2024-08-13
Clicking on **any descendant** of an element with `mv-action` triggers the corresponding action


DmitrySharabin:
Initially found by @LeaVerou.

In many cases, this behavior is not desirable. Especially when one clicks a form element inside an element with `mv-action`.

Testcase: https://codepen.io/dmitrysharabin/pen/eYweRPy

This happens because we run the corresponding action not on the target element itself but on the closest element with a `mv-action` attribute:
https://github.com/mavoweb/mavo/blob/dfe31493f767f1b20b5d34f3ff5c8666d117964b/src/actions.js#L8

LeaVerou:
> This happens because we run the corresponding action not on the target element itself but on the closest element with a `mv-action` attribute:
> 
> https://github.com/mavoweb/mavo/blob/dfe31493f767f1b20b5d34f3ff5c8666d117964b/src/actions.js#L8

That's not a problem, of course we want to run it on the element that actually has an `mv-action` attribute! Otherwise the results would be completely unpredictable.

However, we first need to filter based on `evt.target` and ignore clicks on Mavo controls. I think even a test as simple as `evt.target.closest(".mv-ui")` would be a huge improvement.

DmitrySharabin:
> However, we first need to filter based on `evt.target` and ignore clicks on Mavo controls. I think even a test as simple as `evt.target.closest(".mv-ui")` would be a huge improvement.

This also means that we will ignore generated buttons with both the `mv-ui` class and `mv-action` attribute and the corresponding actions will not be triggered. 

Ignoring `evt.target.closest(".mv-editor")` helps in some cases, but opening, e.g., a built-in color picker, also triggers the action and is not covered by the selector.

Ensure what can be a quick fix here. 🤷‍♂️

LeaVerou:
> > However, we first need to filter based on `evt.target` and ignore clicks on Mavo controls. I think even a test as simple as `evt.target.closest(".mv-ui")` would be a huge improvement.
> 
> This also means that we will ignore generated buttons with both the `mv-ui` class and `mv-action` attribute and the corresponding actions will not be triggered (e.g., the Add list item button won't work).
> 
> Ignoring `evt.target.closest(".mv-editor")` helps in some cases, but opening, e.g., a built-in color picker, also triggers the action and is not covered by the selector.
> 
> Ensure what can be a quick fix here. 🤷‍♂️

[...]

PR #1037, opened 2024-08-13
Ignore clicks on elements inside Mavo controls when triggering actions

DmitrySharabin:
Partially addresses #1036.

Cases, when one clicks a form control inside an `[mv-action]` element, are not addressed by this PR. But I still think this PR is an improvement on the current state.

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -4,11 +4,16 @@ Mavo.attributes.push("mv-action");
 
 let _ = Mavo.Actions = {
 	listener: evt => {
+		// Clicking elements inside these elements should be ignored
+		// if the target element is closer to them than to the [mv-action] element
+		let controlSelector = ".mv-ui, .mv-editor, .mv-popup, .mv-drag-handle";
+
 		let tag = evt.type === "submit"? "form" : ":not(form)";
 		let element = evt.target.closest(tag + "[mv-action]");
+		let control = evt.target.closest(tag + controlSelector);
 
-		if (!element) {
-			return; // Not an action
+		if (!element || control && !control.contains(element)) {
+			return; // Not an action or the click should be ignored
 		}
 
 		let node = Mavo.Node.get(element);
--- a/src/actions.js
+++ b/src/actions.js
@@ -10,7 +10,7 @@ let _ = Mavo.Actions = {
 
 		let tag = evt.type === "submit"? "form" : ":not(form)";
 		let element = evt.target.closest(tag + "[mv-action]");
-		let control = evt.target.closest(tag + controlSelector);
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
Okay, I checked the demos and other apps, and everything seems to work fine with the changes in this PR.

src/actions.js
```diff
@@ -4,11 +4,16 @@ Mavo.attributes.push("mv-action");
 
 let _ = Mavo.Actions = {
 	listener: evt => {
+		// Clicking elements inside these elements should be ignored
+		// if the target element is closer to them than to the [mv-action] element
+		let controlSelector = ".mv-ui, .mv-editor, .mv-popup, .mv-drag-handle";
+
 		let tag = evt.type === "submit"? "form" : ":not(form)";
 		let element = evt.target.closest(tag + "[mv-action]");
+		let control = evt.target.closest(tag + controlSelector);
```
LeaVerou, line 13: This does not intersect the selectors. You'd need something like that to do that:
```suggestion
		let control = evt.target.closest(tag + `:is(${ controlSelector })`);
```

LeaVerou (approved):
LGTM except the thing I mentioned.

I think you could probably fix it for form controls too quite easily.

src/actions.js
DmitrySharabin: Thanks!

DmitrySharabin:
> I think you could probably fix it for form controls too quite easily.

Okay, I added selectors for form controls too. If I miss anything, I think we can add them easily on the way. 🙂

src/actions.js
```diff
@@ -4,11 +4,19 @@ Mavo.attributes.push("mv-action");
...
+			button, label, input, output, select, textarea, meter, progress
+		`;
+
 		let tag = evt.type === "submit"? "form" : ":not(form)";
 		let element = evt.target.closest(tag + "[mv-action]");
+		let control = evt.target.closest(tag + `:is(${ controlSelector })`);
 
-		if (!element) {
-			return; // Not an action
+		if (!element || control && !control.contains(element)) {
```
LeaVerou, line 18: Have you tested what happens if the `mv-action` is *on* these elements? Then it should be respected.

DmitrySharabin: I have. It is respected. The testcase covers such cases: https://codepen.io/dmitrysharabin/pen/eYweRPy

DmitrySharabin:
We should probably add some more elements to the list of selectors that shouldn't trigger actions, like 'a', `audio`, and `video`. Are there any others?

LeaVerou:
> We should probably add some more elements to the list of selectors that shouldn't trigger actions, like 'a', `audio`, and `video`. Are there any others?

`<summary>`

PR #1038, opened 2024-08-14
[actions] Add interactive elements to the list of exceptions

DmitrySharabin:
Otherwise, clicking those elements might produce undesired side effects. For example, in the Stickies demo, following the link to the Open Color website adds a sticky.


LeaVerou (approved):
Approved but don't we have a list of these seomwhere else in the code? I think we need them for the editing UI too

DmitrySharabin:
> Approved but don't we have a list of these seomwhere else in the code? I think we need them for the editing UI too

We have something similar (but not a list) in `elements.js`, and it was my first go-to place. I don't recall having it in any other places. With the changes we made today and yesterday, there is probably a need to store them somewhere in one place.

